PR #12, opened 2025-06-09
Add support for prompts

marcdumontier:
Implement support for prompts (https://modelcontextprotocol.io/docs/concepts/prompts)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,35 @@
 # Changelog
 
+## 0.7.1 (2025-06-20)
+
+### OAuth 2.1 Authentication Framework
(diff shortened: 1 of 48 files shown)


marcdumontier:
added the tests 
needed to make some changes for checking capabilities..because if the server had no tools, but just prompts, it would raise an exception which I didn't think was the correct behaviour

Also made a fix to the tool cache. If you have multiple servers with same tool name, the tool cache would mess things up as it only kept the last one added to the cache.

marcdumontier:
FYI to the update @simonx1

simonx1:
Added in https://github.com/simonx1/ruby-mcp-client/pull/31

PR #158, opened 2026-07-20
Handle interleaved messages on Streamable HTTP POST SSE response streams

simonx1:
## What

Fixes MCP 2025-11-25 compliance of the Streamable HTTP transport's POST SSE response handling. Per [basic/transports — Sending Messages to the Server](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server), when a POST returns `Content-Type: text/event-stream`:

- the stream **MAY** contain JSON-RPC requests and notifications (progress, ping, …) related to the originating request *before* the response — the client **MUST** support this case;
- the server **SHOULD** send a priming event carrying only an event id and empty data;
- events may omit the `event:` field — the SSE default event type is `message`.

The parser previously selected the **first** `event: message` event as the response, which broke all three cases:

| Server behavior (spec-legal) | Old client behavior |
|---|---|
| `notifications/progress` streamed before the tool result | Notification consumed as the response → `call_tool` silently returned `nil`, real result discarded, notification never dispatched |
| Server request (e.g. `ping`) on the POST stream | Never answered — violates the ping MUST-respond requirement |
| Bare `data: {...}` event with no `event:` line | Dropped entirely → `TransportError` + retry re-POST against a server that answered correctly |
| Priming event (`id:` + empty `data:`) | Broke parsing or lost the event id |

## How

`parse_sse_response` now:

1. collects SSE events with the default type `message` when `event:` is omitted, keeping id-only priming events;
2. tracks every event id for resumability;
3. dispatches each interleaved server message through the same dispatcher the GET events stream uses (`dispatch_server_message`: pings answered, server requests routed, notifications delivered to `on_notification`);
4. returns the JSON-RPC response whose `id` matches the originating request — falling back, with a warning, to a lone response whose id does not match (compatibility with servers that mis-echo ids);
5. skips individually malformed events with a warning instead of failing the whole request.

The originating request is threaded through `send_jsonrpc_request` → `parse_response(response, request)` (plain-HTTP transport signature widened, behavior unchanged).

## Tests

- New `spec/lib/mcp_client/server_streamable_http_post_sse_stream_spec.rb` covering all of the above at the WebMock wire level (8 examples, all failing before the fix).
- Two existing tests that encoded the defective behavior updated: the mixed-stream integration test now asserts the response survives a malformed interleaved event (previously asserted the request *fails*), and the id-only-event parser test now asserts priming-event semantics.
- Full suite: 1329 examples, 0 failures; RuboCop clean.

Part of an MCP 2025-11-25 compliance audit of this gem against the official spec (modelcontextprotocol/modelcontextprotocol @ 2025-11-25).

### DIFF
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -153,7 +153,7 @@ module MCPClient
 
       begin
         response = send_http_request(request)
-        parse_response(response)
+        parse_response(response, request)
       rescue MCPClient::Errors::ConnectionError, MCPClient::Errors::TransportError, MCPClient::Errors::ServerError
         raise
       rescue JSON::ParserError => e
@@ -288,7 +288,7 @@ module MCPClient
     # @param response [Faraday::Response] the HTTP response
     # @return [Hash] the parsed result
     # @raise [NotImplementedError] if not implemented by concrete transport
-    def parse_response(response)
+    def parse_response(response, _request = nil)
       raise NotImplementedError, 'Subclass must implement parse_response'
     end
   end
--- a/lib/mcp_client/server_http/json_rpc_transport.rb
+++ b/lib/mcp_client/server_http/json_rpc_transport.rb
@@ -12,10 +12,11 @@ module MCPClient
 
       # Parse an HTTP JSON-RPC response
       # @param response [Faraday::Response] the HTTP response
+      # @param _request [Hash, nil] the originating JSON-RPC request (unused)
       # @return [Hash] the parsed result
       # @raise [MCPClient::Errors::TransportError] if parsing fails
       # @raise [MCPClient::Errors::ServerError] if the response contains an error
-      def parse_response(response)
+      def parse_response(response, _request = nil)
         body = response.body.strip
         data = JSON.parse(body)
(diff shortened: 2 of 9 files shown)


simonx1:
External verification (Codex) reviewed this diff. Outcome:

**Fixed in follow-up commits:** `@last_event_id` writes on the POST path are now synchronized on the transport monitor.

**Deferred (tracked for the SEP-1699 resumability PR, which reworks this code):**
- The POST body is still buffered by Faraday before parsing, so interleaved server messages are dispatched at stream completion rather than incrementally. A server that blocks its response on an interleaved ping's pong would stall to the read timeout. Incremental `on_data` dispatch lands together with the retry/Last-Event-ID/GET-resumption fixes.
- `Last-Event-ID` is currently attached to POSTs and the `retry:` field is ignored — both are the subject of the SEP-1699 PR.

**Kept by design:** the lone-response fallback for a mismatched id. It only fires when *no* response matches the request id and exactly one response-shaped message exists on the stream, it logs a warning, and it preserves compatibility with servers that mis-echo ids. Without it, such servers would turn every request into a retryable transport error.

PR #159, opened 2026-07-20
Fix elicitation compliance: JSON-RPC responses, mode negotiation, error semantics

simonx1:
## What

Four elicitation fixes against [MCP 2025-11-25 client/elicitation](https://modelcontextprotocol.io/specification/2025-11-25/client/elicitation):

### 1. Streamable HTTP: reply with a real JSON-RPC response (MUST)

The transport answered a server's `elicitation/create` request with an invented message: `{"jsonrpc":"2.0","method":"elicitation/response","params":{"elicitationId":<the JSON-RPC id>,...}}` — no `id`, no `result`, and a method that exists nowhere in the spec. A compliant server never receives a response to its request (it hangs until timeout) and additionally gets an unknown-method notification it can only ignore. The sibling `roots/list` and `sampling/createMessage` reply paths in the same class already POSTed proper responses; elicitation now does the same: `{"jsonrpc":"2.0","id":<request id>,"result":{...ElicitResult...}}`.

### 2. Mode negotiation (MUST)

> "Server sends an `elicitation/create` request with a mode not declared in client capabilities: `-32602` (Invalid params)"

Previously **any** unknown mode string was silently processed as form mode, and the declared capability was the bare `'elicitation' => {}` — which the spec defines as *form-only*, making the gem's implemented URL-mode handling unreachable via compliant servers. Now:
- capability declares both implemented modes: `{'form' => {}, 'url' => {}}`;
- undeclared/unknown modes are rejected with `-32602` on every transport.

### 3. Truthful error semantics

Internal failures were misreported as user actions — a handler exception or missing handler produced `{'action' => 'decline'}`, which the spec defines as "User explicitly declined the request" and tells servers to act on. Now: handler exception → `-32603` Internal error; no handler → `-32601`; error-shaped results flow through each transport's `send_elicitation_response` into a JSON-RPC error response, mirroring the existing sampling error path.

### 4. Validation (SHOULD)

> "Clients SHOULD validate all responses against the provided schema"

Validation was warn-only — content violating the `requestedSchema` was logged and then transmitted anyway. Now non-conforming form accepts return `-32603` with the validation errors instead of sending invalid content, and the validator checks the spec's string `format` constraints (`email`, `uri`, `date`, `date-time`) which were previously ignored. URL-mode accepts omit `content` per the `ElicitResult` schema ("Omitted for out-of-band mode responses").

## Tests

- New `spec/lib/mcp_client/elicitation_compliance_spec.rb` (16 examples: client-level error codes, mode rejection, URL-mode content omission, validation enforcement, format checks, capability shape, and wire-level Streamable HTTP request/response + error cycles). 13 failed before the fix.
- Existing tests that encoded the old behavior updated (auto-decline expectations, `elicitation/response` wire stubs, capability shape).
- Full suite: 1336 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -938,19 +938,21 @@ module MCPClient
     # @param params [Hash] the elicitation parameters
     # @return [Hash] the elicitation response
     def handle_elicitation_request(_request_id, params)
+      mode = params['mode'] || 'form'
+      # MCP 2025-11-25: requests with a mode not declared in client
+      # capabilities MUST be rejected with -32602 (Invalid params). This check
+      # precedes everything else — an undeclared mode is -32602 even when no
+      # handler is configured.
+      unless SUPPORTED_ELICITATION_MODES.include?(mode)
+        @logger.warn("Rejecting elicitation request with unsupported mode '#{mode}'")
+        return jsonrpc_error_result(-32_602, "Elicitation mode '#{mode}' is not supported")
+      end
+
       # Without a handler there is no user to interact with: answer with a
       # JSON-RPC error rather than fabricating a user "decline".
       unless @elicitation_handler
         @logger.warn('Received elicitation request but no elicitation handler is configured')
         return jsonrpc_error_result(-32_601, 'Elicitation not supported: no elicitation handler configured')
-      end
-
-      mode = params['mode'] || 'form'
-      # MCP 2025-11-25: requests with a mode not declared in client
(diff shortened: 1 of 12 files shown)


simonx1:
External verification (Codex) reviewed this diff; the three substantive findings (undeclared-mode check must precede the missing-handler check, scalar content must not be transmitted as ElicitResult.content, symbol/mixed-key handler results bypassing URL-mode content omission) are fixed in follow-up commits with regression tests. The capability-declaration observations are addressed by the stacked PR #160.

PR #160, opened 2026-07-20
Derive declared client capabilities from registered handlers

simonx1:
## What

Fixes hardcoded client capability declarations that violated the MCP 2025-11-25 lifecycle MUSTs ([basic/lifecycle — Operation](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#operation): "Only use capabilities that were successfully negotiated"; [client/sampling](https://modelcontextprotocol.io/specification/2025-11-25/client/sampling#capabilities) / roots / elicitation: "Clients that support X **MUST** declare the X capability during initialization"):

**Under-declaration (Streamable HTTP):** `HttpTransportBase#initialization_params` sent `'capabilities' => {}` even though `ServerStreamableHTTP` fully implements `elicitation/create`, `roots/list`, and `sampling/createMessage`. A conforming server never sends those requests to a client that didn't declare them — the gem's documented elicitation/sampling/roots support was unreachable on its flagship transport. Worse, `Client#roots=` still sent `notifications/roots/list_changed` to those sessions, using a capability that was never negotiated.

**Over-declaration (stdio/SSE):** `elicitation`, `roots`, and `sampling` were declared unconditionally, even when the host registered no handler — a server then legitimately using the negotiated capability got an error back.

## How

- Declared capabilities are now derived in one shared place (`JsonRpcCommon#client_capabilities`) from the server-request callbacks actually registered before connect. The divergent `HttpTransportBase` override is deleted, so all four transports share the logic. Plain `ServerHTTP` (no server-request channel) naturally continues to declare `{}`.
- `Client` registers the elicitation/sampling callbacks only when the host configured the corresponding handler. Roots stays always-registered: the client implements `roots/list` and change notifications regardless of the current (possibly empty) roots list.
- `notify_roots_changed` skips transports without a roots request channel, so `notifications/roots/list_changed` is only sent where the capability was declared.

## Tests

New `spec/lib/mcp_client/client_capabilities_spec.rb` (7 examples: per-callback derivation on stdio, wire-level declaration on Streamable HTTP, Client-level conditional registration, and roots-notification gating; 6 failed before the fix). Existing tests locking the old always-register/hardcoded behavior updated. Full suite: 1344 examples, 0 failures; RuboCop clean.

> **Note:** stacked on #159 (`fix/elicitation-compliance`) — both touch the capability declaration in `json_rpc_common.rb`. Merge #159 first; GitHub will retarget this PR to `main` automatically when the base branch is deleted.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -6,6 +6,10 @@ module MCPClient
   # MCP Client for integrating with the Model Context Protocol
   # This is the main entry point for using MCP tools
   class Client
+    # Elicitation modes implemented by this client (MCP 2025-11-25).
+    # Requests with a mode outside this set are rejected with -32602.
+    SUPPORTED_ELICITATION_MODES = %w[form url].freeze
+
     # @!attribute [r] servers
     #   @return [Array<MCPClient::ServerBase>] list of servers
     # @!attribute [r] tool_cache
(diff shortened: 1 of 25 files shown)


simonx1:
External verification (Codex) reviewed this diff. On the mode-advertising observation: declaring both elicitation modes for any registered callback is deliberate — the transport callback receives the full request params (including `mode`) and forwards every mode to the host handler, and the client-level handler implements both modes; a mode-restricted declaration API can follow if a form-only host materializes. The registration/connect race requires registering callbacks concurrently with connect, which the documented flow (register before connect, as MCPClient::Client does) avoids.

PR #161, opened 2026-07-20
Enforce protocol version negotiation: disconnect on unsupported versions

simonx1:
## What

Implements the version-negotiation MUST from the MCP lifecycle ([basic/lifecycle — Version Negotiation](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#version-negotiation), `InitializeResult.protocolVersion` in the official schema):

> "The version of the Model Context Protocol that the server wants to use. This may not match the version that the client requested. If the client cannot support this version, it **MUST** disconnect."

No transport performed any check:

- **stdio** and **legacy SSE** never even read `result['protocolVersion']` — the negotiated version was discarded with no public accessor, so a host couldn't perform the check either.
- **HTTP / Streamable HTTP** stored it without comparison and then dutifully echoed the unvalidated value in the `MCP-Protocol-Version` header.

A server negotiating e.g. `1999-01-01` (or omitting the field entirely) silently continued the session under 2025-11-25 semantics.

## How

- New `MCPClient::SUPPORTED_PROTOCOL_VERSIONS = %w[2025-11-25 2025-06-18 2025-03-26 2024-11-05]` in `version.rb`.
- Shared `JsonRpcCommon#validate_protocol_version!`: returns the negotiated version when supported; otherwise runs the transport's `cleanup` (disconnect) and raises `ConnectionError` naming the offending version.
- Called from every transport's `perform_initialize`; stdio and SSE now also record `@protocol_version` (needed for the `MCP-Protocol-Version` header work on the legacy SSE transport, tracked separately).

## Tests

- New `spec/lib/mcp_client/protocol_version_negotiation_spec.rb` (8 examples across Streamable HTTP wire level, stdio, and SSE: unsupported version → disconnect+raise, missing version → raise, supported older version → accepted and recorded). 7 failed before the fix.
- Existing stubs that answered `initialize` with an empty result (non-compliant per the `InitializeResult` schema, which requires `protocolVersion`) now include one — including the fake stdio filesystem server used by the integration suite.
- Full suite: 1328 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -137,9 +137,10 @@ module MCPClient
       result = send_jsonrpc_request(json_rpc_request)
       return unless result.is_a?(Hash)
 
+      # Disconnects if the server negotiated a version we cannot speak.
+      @protocol_version = validate_protocol_version!(result)
       @server_info = result['serverInfo']
       @capabilities = result['capabilities']
-      @protocol_version = result['protocolVersion']
     end
 
     # Send a JSON-RPC request to the server and wait for result
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -135,7 +135,10 @@ module MCPClient
       @logger.debug("Performing initialize RPC: #{json_rpc_request}")
 
       result = send_jsonrpc_request(json_rpc_request)
-      return unless result.is_a?(Hash)
+      unless result.is_a?(Hash)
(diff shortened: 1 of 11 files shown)


simonx1:
External verification (Codex) reviewed this diff; the substantive finding (a non-object initialize result bypassed version validation via an early return on the HTTP and SSE paths) is fixed in a follow-up commit with a regression test. The four-revision support list is a documented judgment call; the connect-race and mid-handshake session-cleanup observations are pre-existing transport behaviors noted for follow-up.

PR #162, opened 2026-07-20
Fix Streamable HTTP session lifecycle: 404 re-initialization and session ID acceptance

simonx1:
## What

Implements two session-management MUSTs from [MCP 2025-11-25 basic/transports — Session Management](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management):

### 1. Recover from session expiry (HTTP 404)

> "When a client receives HTTP 404 in response to a request containing an `MCP-Session-Id`, it **MUST** start a new session by sending a new `InitializeRequest` without a session ID attached."

Previously a 404 was mapped to a generic non-retryable `ServerError` while the stale `@session_id` was retained and the connection stayed "established" — every subsequent request re-sent the dead session id and failed identically, forever. Since servers **MAY** terminate sessions at any time (and then **MUST** answer with 404), any long-lived connection to a session-expiring server was permanently broken until the host manually called `cleanup` + `connect`.

`send_http_request` now detects `404` + session-present, clears the session state (`@session_id`, `@last_event_id`), re-runs the transport's initialize handshake (which sends no session header and emits `notifications/initialized`), and resends the original request once. A guard flag prevents loops when the fresh session also 404s — that raises the normal `ServerError`.

### 2. Accept spec-valid session IDs

> "The session ID ... MUST only contain visible ASCII characters (ranging from 0x21 to 0x7E)" — with "a securely generated UUID, a JWT, or a cryptographic hash" as the spec's own examples.

The client-side check was `/\A[a-zA-Z0-9\-_]{8,128}\z/` — rejecting JWTs (`.`), base64 ids (`+`, `/`, `=`), and anything shorter than 8 chars. A rejected id was silently dropped (warning log only), after which the client sent **no** `Mcp-Session-Id` on any subsequent request — violating the MUST to echo it and drawing HTTP 400s from stateful servers. Now validates against the spec charset (`/\A[\x21-\x7E]{1,512}\z/`, generous length cap).

## Tests

- New `spec/lib/mcp_client/streamable_session_lifecycle_spec.rb` (4 examples: JWT/base64/short-id acceptance, non-visible-ASCII rejection, wire-level 404 → re-initialize-without-session-header → resend-with-new-session flow, and no-loop-on-repeated-404). 3 failed before the fix.
- Existing per-character rejection lists updated to the spec charset (entries like `session.id`, `session+id`, `short` are spec-valid and now live in the accepted list; space/control/non-ASCII remain rejected).
- Full suite: 1304 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -175,6 +175,12 @@ module MCPClient
     # @raise [MCPClient::Errors::ConnectionError] if connection fails
     def send_http_request(request)
       conn = http_connection
+      # Capture the session id this request goes out with — the value
+      # apply_request_headers attaches — so a later 404 is attributed to the
+      # id that actually accompanied the request, not to whatever @session_id
+      # holds by 404-handling time (another caller may have completed a
+      # restart in between, and its fresh session must not be re-initialized).
+      sent_session_id = @mutex.synchronize { @session_id }
 
       begin
         response = conn.post(@endpoint) do |req|
@@ -185,8 +191,8 @@ module MCPClient
         # MCP 2025-11-25 session management: HTTP 404 for a request carrying
         # Mcp-Session-Id means the session expired — the client MUST start a
         # new session with a fresh InitializeRequest (without a session ID).
-        if response.status == 404 && (expired_session = expired_session_for_restart)
-          return restart_session_and_resend(request, expired_session)
+        if response.status == 404 && session_restart_applicable?(sent_session_id)
+          return restart_session_and_resend(request, sent_session_id)
         end
 
         handle_http_error_response(response) unless response.success?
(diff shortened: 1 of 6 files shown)


simonx1:
External verification (Codex) reviewed this diff; fixed in a follow-up commit: 404 recovery now also works under user-configured raise_error middleware (previously bypassed into a retryable TransportError with the stale session), the restart is serialized on the transport monitor, and the session-ID length cap is 4096 and documented as header hygiene rather than spec grammar. The generation-tagged single-flight restart and old-events-stream cursor tagging are noted as follow-up hardening beyond this PR's compliance scope.

simonx1:
External delta verification (Codex) re-reviewed the follow-up; the remaining race is fixed: the restart now captures the expired session id and rechecks it under the monitor, so a caller arriving after another caller's restart resends against the fresh session instead of initializing a third one. The 4096-char cap stays as documented header hygiene.

simonx1:
External convergence review (Codex) flagged that the expired id was read after the 404 rather than at send time; fixed with a regression test — the session id attached to each request is captured under the monitor when the request is sent, and 404 recovery compares against that, so a caller whose 404 races a completed restart resends instead of re-initializing.

simonx1:
Final convergence round (Codex): the header/capture race is closed — the Mcp-Session-Id header sent on the wire is now reconciled to the id captured for 404 attribution, with a mid-flight-restart regression test. Suite green (1309 examples), RuboCop clean.

PR #163, opened 2026-07-20
Handle OAuth WWW-Authenticate challenges: 401 discovery, scope selection, 403 step-up

simonx1:
## What

Three authorization fixes against [MCP 2025-11-25 basic/authorization](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization):

### 1. Make 401 `WWW-Authenticate` discovery reachable (MUST)

> "MCP clients **MUST** be able to parse `WWW-Authenticate` headers and respond appropriately to HTTP 401 Unauthorized responses" — using "the resource metadata URL from the parsed `WWW-Authenticate` headers when present".

`OAuthProvider#handle_unauthorized_response` implements the parsing correctly, but the only transport wiring that called it lived behind `Faraday::UnauthorizedError`/`ForbiddenError` rescues — exceptions Faraday only raises when the `:raise_error` middleware is registered, which no transport does. On the actual default path, `handle_http_error_response` raised a plain `ConnectionError` with the header discarded, so a server advertising its Protected Resource Metadata **only** via the header (one of the two mechanisms the spec lets servers choose) could never be discovered. The 401/403 error path now passes the response to the OAuth provider (failures logged, never masking the original error) before raising.

### 2. Spec-compliant scope selection

> "Clients **MUST** treat the scopes provided in the challenge as authoritative for satisfying the current request." Fallback (SHOULD): "use all scopes defined in `scopes_supported` from the Protected Resource Metadata document, omitting the `scope` parameter if `scopes_supported` is undefined."

Neither source existed: the challenge's `scope` parameter was never parsed, and `ResourceMetadata` dropped `scopes_supported` entirely (a host could never see it through this library). Now: the provider captures `challenge_scope` from every challenge (exposed publicly), `ResourceMetadata` parses/round-trips `scopes_supported`, and authorization/registration resolve scope by the spec's priority — challenge scope → configured scope (`:all` keeps its documented AS-metadata behavior) → PRM `scopes_supported` → omit.

### 3. SEP-835 step-up authorization enabled

> "Clients **SHOULD** respond to these errors by requesting a new access token with an increased set of scopes via a step-up authorization flow..."

A 403 `insufficient_scope` challenge was swallowed into `"Authorization failed: HTTP 403"` — the required scopes were unknowable, making step-up impossible for both the library and the host. Now it raises the new `MCPClient::Errors::InsufficientScopeError` (subclass of `ConnectionError`, so existing rescues keep working) exposing `scope` and `error_description`; combined with the captured `challenge_scope`, the next `start_authorization_flow` automatically requests the challenged scopes.

## Tests

New `spec/lib/mcp_client/oauth_challenge_handling_spec.rb` (9 examples: wire-level 401 → provider handoff, 403 insufficient_scope → typed error with scopes, challenge-scope capture with/without resource_metadata, full scope-resolution priority ladder, RFC 9728 `scopes_supported` round-trip). All 9 failed before the fix. Full suite: 1329 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/auth.rb
+++ b/lib/mcp_client/auth.rb
@@ -290,13 +290,16 @@ module MCPClient
 
     # Protected resource metadata for authorization server discovery
     class ResourceMetadata
-      attr_reader :resource, :authorization_servers
+      attr_reader :resource, :authorization_servers, :scopes_supported
 
       # @param resource [String] Resource server identifier
       # @param authorization_servers [Array<String>] List of authorization server URLs
-      def initialize(resource:, authorization_servers:)
+      # @param scopes_supported [Array<String>, nil] Scopes the resource supports (RFC 9728);
+      #   per MCP, the default scope set when a challenge provides none
+      def initialize(resource:, authorization_servers:, scopes_supported: nil)
         @resource = resource
         @authorization_servers = authorization_servers
+        @scopes_supported = scopes_supported
       end
 
       # Convert to hash
@@ -304,8 +307,9 @@ module MCPClient
       def to_h
         {
           resource: @resource,
-          authorization_servers: @authorization_servers
-        }
+          authorization_servers: @authorization_servers,
+          scopes_supported: @scopes_supported
+        }.compact
       end
 
       # Create resource metadata from hash
@@ -314,7 +318,8 @@ module MCPClient
       def self.from_h(data)
         new(
           resource: data[:resource] || data['resource'],
-          authorization_servers: data[:authorization_servers] || data['authorization_servers']
+          authorization_servers: data[:authorization_servers] || data['authorization_servers'],
+          scopes_supported: data[:scopes_supported] || data['scopes_supported']
         )
       end
     end
--- a/lib/mcp_client/auth/oauth_provider.rb
+++ b/lib/mcp_client/auth/oauth_provider.rb
@@ -11,6 +11,18 @@ module MCPClient
     # Handles the complete OAuth flow including server discovery, client registration,
     # authorization, token exchange, and refresh
     class OAuthProvider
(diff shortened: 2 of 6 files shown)


simonx1:
External verification (Codex) reviewed this diff; fixed in a follow-up commit: the raise_error-middleware exception path now routes through the same challenge pipeline (Hash responses normalized; typed InsufficientScopeError raised there too), challenge parsing is Bearer-anchored with token boundaries (no more Basic/prefixed-token/prefixed-param false positives), a scope-less challenge clears the previously stored scope, :all falls back to PRM scopes when the AS list is empty, and the challenge-advertised metadata URL is retained for discovery retry. Per-request challenge-scope snapshotting is noted as follow-up hardening.

simonx1:
External delta verification (Codex) re-reviewed the follow-up; both remaining findings fixed with regression tests: challenge parameters are now parsed only within the Bearer challenge's own segment (mixed-scheme headers no longer misclassify, and the error token must match insufficient_scope exactly), and a challenge-advertised metadata URL is authoritative — it bypasses cached metadata and is fetched strictly with no well-known fallback, per the discovery MUST.

simonx1:
External convergence review (Codex) flagged that the provider still parsed challenge parameters from the whole header; fixed with regression tests — handle_unauthorized_response now extracts scope and resource_metadata only from the Bearer challenge's own segment, so parameters of other schemes cannot drive Bearer discovery or scope selection.

simonx1:
Final convergence round (Codex): the quoted-string anchoring case is closed — both Bearer-segment extractors mask quoted values before locating the scheme token, with regression tests for embedded-Bearer headers on the transport and provider paths. Suite green (1348 examples), RuboCop clean.

PR #165, opened 2026-07-20
Refuse authorization when PKCE support cannot be verified

simonx1:
## What

Implements the PKCE verification MUST from [MCP 2025-11-25 basic/authorization — Security Considerations](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#security-considerations):

> "If `code_challenge_methods_supported` is absent, the authorization server does not support PKCE and MCP clients **MUST** refuse to proceed." (The OIDC-discovery variant of the same rule is equally explicit.)

`verify_pkce_support!` already refused when the field was present without `S256`, but when the field was **absent entirely** it logged a warning and continued the flow. That defeats the downgrade protection the requirement exists for: an AS that doesn't understand `code_challenge` silently ignores it, leaving the authorization code unprotected. Both discovery document types flow through this same validation, so the fix covers both MUSTs.

Absence now raises `ConnectionError`, mirroring the existing S256-missing branch.

## Tests

The warn-and-proceed test now asserts the refusal; the `scope: :all` fixtures gained `code_challenge_methods_supported: ['S256']` (they modeled a non-PKCE server, which is exactly what must now be rejected). Full suite: 1320 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/auth/oauth_provider.rb
+++ b/lib/mcp_client/auth/oauth_provider.rb
@@ -444,17 +444,18 @@ module MCPClient
       end
 
       # Verify the authorization server supports PKCE S256 (RFC 8414 / MCP).
-      # Refuses when S256 is explicitly not offered; warns (and proceeds, since
-      # the client always uses S256) when the field is not advertised at all.
+      # Per MCP 2025-11-25, "If code_challenge_methods_supported is absent,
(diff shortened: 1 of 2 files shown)


simonx1:
External verification (Codex) reviewed this diff: verdict correct with no findings; test coverage judged adequate.

PR #167, opened 2026-07-20
Attach OAuth authorization to every Streamable HTTP request

simonx1:
## What

Implements the per-request token MUST from [MCP 2025-11-25 basic/authorization — Access Token Usage](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#access-token-usage):

> "Note that authorization **MUST** be included in every HTTP request from client to server, even if they are part of the same logical session."

The Bearer token was only applied on the primary POST path. Four request paths hand-assembled their headers and sent **no** `Authorization`, so against an OAuth-protected server:

| Path | Consequence |
|---|---|
| GET events stream | 401 → events thread reconnect-loops forever; server-initiated requests/notifications never arrive |
| Session-termination DELETE (shared with plain HTTP transport) | 401 → session never terminated server-side |
| Pong responses to server pings | 401 → server sees its pings time out |
| POSTed responses to server-initiated elicitation/roots/sampling requests | 401 → server sees its requests time out |

All four now apply `@oauth_provider&.apply_authorization(req)` alongside the session and protocol-version headers.

## Tests

New `spec/lib/mcp_client/oauth_token_all_requests_spec.rb` — 5 wire-level examples asserting the Authorization header on the GET stream, both transports' DELETE, pong POSTs, and server-request response POSTs. All 5 failed before the fix. Full suite: 1325 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -59,6 +59,8 @@ module MCPClient
           @headers.each { |k, v| req.headers[k] = v }
           req.headers['Mcp-Session-Id'] = @session_id
           req.headers['Mcp-Protocol-Version'] = @protocol_version if @protocol_version
+          # MCP: authorization MUST be included in every HTTP request
+          @oauth_provider&.apply_authorization(req)
         end
 
         if response.success?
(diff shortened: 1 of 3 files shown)


simonx1:
External verification (Codex) reviewed this diff; no findings on the request-coverage change itself. The noted token-refresh race in OAuthProvider#access_token under concurrent background requests is pre-existing provider behavior (now simply more exercised) and is tracked as follow-up hardening — a refresh mutex belongs in the provider, which this PR deliberately does not touch.

PR #168, opened 2026-07-20
Implement SEP-1699 resumability: retry directive, GET-based resumption, cursor placement

simonx1:
## What

Reworks Streamable HTTP reconnection per [MCP 2025-11-25 basic/transports — Resumability and Redelivery](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#resumability-and-redelivery) (SEP-1699 + its clarifications):

### 1. The SSE `retry` directive is honored (MUST)

> "The client **MUST** respect the `retry` field, waiting the given number of milliseconds before attempting to reconnect."

It was parsed, logged, and ignored — the events loop slept its own exponential backoff (1s→30s) regardless. A server directing `retry: 60000` was reconnected to up to 60× too often; `retry: 250` fast polling waited 4–120× too long. Both the GET stream parser and the POST SSE parser now record it, and every reconnect sleep consults it.

### 2. Cursor placement fixed (SHOULD)

> "If the client wishes to resume after a disconnection ... it **SHOULD** issue an HTTP GET to the MCP endpoint, and include the `Last-Event-ID` header." — "Resumption is always via HTTP GET with `Last-Event-ID`," regardless of how the original stream was initiated.

The implementation had it inverted: `Last-Event-ID` was attached to **every subsequent JSON-RPC POST** (meaningless there, potentially confusing servers) and **never** to GET reconnects — so replay of missed messages could never happen. Now the cursor rides only on the GET events request.

### 3. The polling pattern works without duplicate execution

Per SEP-1699 a server **MAY** close a stream at any time (priming event, then close, response later). Previously that raised a `TransportError`, and `with_retry` re-POSTed the identical JSON-RPC request — re-executing a non-idempotent `tools/call`. Now, when a POST stream ends with a cursor but no response, the client resumes via the GET events stream (which carries the cursor), waits — bounded by `read_timeout` — for the replayed response routed back to the caller, and returns it. If resumption fails, it raises a **non-retryable** `ServerError`, because the request may already be executing server-side.

**Honest scope note:** interleaved messages on a still-open POST stream are dispatched when the stream completes (Faraday buffers POST bodies); the SEP-1699-sanctioned server behavior — closing the stream — is now fully handled via GET resumption, which *is* incremental. Fully incremental POST-body parsing remains a possible future refinement.

## Tests

New `spec/lib/mcp_client/streamable_resumability_spec.rb` (8 examples: no cursor on POSTs, cursor on GET, retry recorded from both stream types, retry-driven reconnect delay, wire-level resume-via-GET returning the replayed response with exactly one POST, and non-retryable failure without re-POST). All 8 failed before the fix. Existing tests asserting the old cursor-on-POST behavior updated to assert the SEP-1699 placement. Full suite: 1337 examples, 0 failures; RuboCop clean.

> Stacked on #158 (`fix/streamable-http-post-sse-stream`) — builds on its SSE parser rework. Merge #158 first.

### DIFF
--- a/lib/mcp_client/server_streamable_http.rb
+++ b/lib/mcp_client/server_streamable_http.rb
@@ -728,22 +728,20 @@ module MCPClient
     end
 
     # Wait for a response replayed after the POST stream was closed before
-    # delivering it (SEP-1699 polling pattern). Issues a dedicated HTTP GET
-    # carrying the disconnected stream's Last-Event-ID cursor — the general
-    # events stream (opened without that cursor) cannot receive the replay.
+    # delivering it (SEP-1699 polling pattern). A dedicated worker issues
+    # HTTP GETs carrying the disconnected stream's Last-Event-ID cursor — the
+    # general events stream (opened without that cursor) cannot receive the
+    # replay.
     # @param request_id [Integer, String] id of the outstanding request
     # @param cursor [String] last event id received on the closed stream
+    # @param retry_ms [Integer, nil] retry directive received on the closed
+    #   stream itself (not the shared events-stream directive)
     # @return [Hash, nil] the replayed JSON-RPC response, or nil on timeout
-    def resume_response_via_get(request_id, cursor)
+    def resume_response_via_get(request_id, cursor, retry_ms = nil)
       queue = Thread::Queue.new
       @mutex.synchronize { @pending_stream_responses[request_id] = queue }
 
-      # SEP-1699: the client MUST respect the server's retry directive before
-      # attempting to reconnect.
-      retry_ms = @sse_retry_ms
-      sleep(retry_ms / 1000.0) if retry_ms&.positive?
-
-      resume_thread = Thread.new { issue_resumption_get(cursor) }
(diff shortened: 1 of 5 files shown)


simonx1:
External verification (Codex) reviewed this diff. All three findings fixed in a follow-up commit with regression tests: resumption now issues a dedicated GET carrying the closed stream's own Last-Event-ID cursor (the live general events stream, opened without the cursor, could never receive the replay — and per-stream cursor selection replaces the shared @last_event_id read), the server's retry directive is honored before the resumption reconnect, and retry: 0 is accepted as a valid immediate-reconnect directive on both parsers.

simonx1:
External delta verification (Codex) re-reviewed the follow-up; all three findings fixed with regression tests: resumption now loops until the deadline, tracking id:/retry: fields on the resumed stream itself (so a re-closed resumed connection reconnects with the fresh cursor after the directed delay), events split on both CRLF and LF separators, and the resumption delay is taken from the originating POST stream's own retry directive rather than the shared ivar.

PR #169, opened 2026-07-20
Validate root URIs (file:// MUST) and expose Root _meta

simonx1:
## What

The MCP 2025-11-25 roots spec (`client/roots.mdx`, Data Types — Root) states:

> `uri`: Unique identifier for the root. This **MUST** be a `file://` URI in the current specification.

and, under Security Considerations:

> Clients **MUST**: ... Validate all root URIs to prevent path traversal

Previously `MCPClient::Root` accepted any uri value (including `nil`, `https://` URIs, or bare paths) and the client emitted it verbatim in `roots/list` responses, producing spec-invalid Roots on the wire.

- `Root#initialize` now raises `ArgumentError` unless the uri is a parseable `file://` URI (scheme compared case-insensitively) whose path contains no `..` traversal segments. Since `Client#initialize` and `Client#roots=` both funnel through `normalize_roots` → `Root.new`/`Root.from_json`, both entry points are covered with no client changes.
- `Root` also gains the optional `_meta` field declared in `schema.ts` (`_meta?: { [key: string]: unknown }`): a `meta:` kwarg, parsed from `_meta` in `from_json` (string or symbol keys), emitted in `to_h` only when present, and included in `==`/`hash`. It was previously silently dropped, so hosts could never attach metadata to roots.

## Tests

- New `spec/lib/mcp_client/roots_validation_spec.rb` (24 examples): rejects `http://`/`https://`, relative and schemeless paths, `nil`/non-string/unparseable uris, and `..` traversal; accepts `file:///path` (including uppercase scheme); `_meta` round-trips through `from_json`/`to_h` and appears in the `roots/list` response; `Client.new(roots: [...])` and `Client#roots=` raise on invalid roots.
- Full suite: 1344 examples, 0 failures. RuboCop: 111 files inspected, no offenses.

### DIFF
--- a/lib/mcp_client/root.rb
+++ b/lib/mcp_client/root.rb
@@ -17,6 +17,9 @@ module MCPClient
     # @raise [ArgumentError] if uri is not a valid file:// URI or contains '..' path segments
     def initialize(uri:, name: nil, meta: nil)
       validate_uri!(uri)
+      # Root._meta is an object of arbitrary keys per the schema
+      raise ArgumentError, "Root _meta must be a Hash, got #{meta.class}" if meta && !meta.is_a?(Hash)
+
       @uri = uri
       @name = name
       @meta = meta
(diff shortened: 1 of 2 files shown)


simonx1:
External verification (Codex) reviewed this diff. All three findings fixed in a follow-up commit with regression tests: percent-encoded traversal segments are decoded before the '..' check, the uri must start with the literal file:// form (file:relative and file:/path rejected), and Root _meta must be an object.

PR #170, opened 2026-07-20
Parse and expose 2025-11-25 metadata fields (icons, title, _meta) on model classes

simonx1:
## What

MCP 2025-11-25 added `icons` (SEP-973) to Tool, Prompt, Resource, ResourceTemplate, and ResourceLink, and reserves `_meta` on schema objects so servers can attach metadata for clients to pass through. The typed model classes silently dropped these wire fields — and `Prompt` additionally dropped `title` — which is interop-affecting: because `from_json` is the mandatory parse path for `tools/list`, `prompts/list`, `resources/list`, `resources/templates/list`, `resources/read`, and `resource_link` content blocks, hosts using this library could never display a tool/prompt/resource icon or a prompt title, and could never read server-attached `_meta`.

This PR parses and exposes the dropped fields, mirroring the exact field names in `schema/2025-11-25/schema.ts`:

- `Tool`: `icons`, `_meta` (exposed as `#icons` / `#meta`; both string and symbol keys, matching the rest of `Tool.from_json`)
- `Prompt`: `title`, `icons`, `_meta`
- `Resource`, `ResourceTemplate`, `ResourceLink`: `icons`, `_meta`
- `ResourceContent`: `_meta` (ResourceContents does not carry `icons` in the schema, so only `_meta` is added)

All new constructor keyword arguments default to `nil`, so existing callers are unaffected. No fields beyond those in the official schema were introduced.

## Tests

- New `spec/lib/mcp_client/model_metadata_fidelity_spec.rb` (written first, red before the fix): one example per class asserting a fully-populated wire hash round-trips `icons`/`title`/`_meta` through `from_json`.
- Full suite: `bundle exec rspec` — 1326 examples, 0 failures.
- `bundle exec rubocop` — 111 files inspected, no offenses detected.

### DIFF
--- a/lib/mcp_client/prompt.rb
+++ b/lib/mcp_client/prompt.rb
@@ -5,23 +5,35 @@ module MCPClient
   class Prompt
     # @!attribute [r] name
     #   @return [String] the name of the prompt
+    # @!attribute [r] title
+    #   @return [String, nil] optional human-readable name of the prompt for display purposes
     # @!attribute [r] description
     #   @return [String] the description of the prompt
     # @!attribute [r] arguments
     #   @return [Hash] the JSON arguments for the prompt
+    # @!attribute [r] icons
+    #   @return [Array<Hash>, nil] optional icons for display in user interfaces (MCP 2025-11-25, SEP-973)
+    # @!attribute [r] meta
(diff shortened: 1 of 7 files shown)


simonx1:
External verification (Codex) reviewed this diff: verdict correct with no findings (field mappings verified against the official schema.ts); test coverage judged adequate.

PR #171, opened 2026-07-20
Enforce UTF-8 on stdio pipes and implement the spec shutdown sequence

simonx1:
## What

Fixes two verified findings against the MCP 2025-11-25 spec in the stdio transport, plus one reader-thread robustness fix.

**1. UTF-8 is now enforced on the subprocess pipes** (`basic/transports.mdx`):

> MCP uses JSON-RPC to encode messages. JSON-RPC messages **MUST** be UTF-8 encoded.

`connect` now pins `@stdin`/`@stdout`/`@stderr` to UTF-8 (`set_encoding(Encoding::UTF_8)`) right after `Open3.popen3`. Previously the pipes inherited `Encoding.default_external`, so under a non-UTF-8 locale (e.g. `LANG=C` in containers, systemd units, CI) a fully spec-compliant response containing any non-ASCII byte raised `Encoding::InvalidByteSequenceError` inside the reader thread, silently killing it and causing every subsequent request to time out. `@stderr` is pinned too, since the server "**MAY** write UTF-8 strings to its standard error".

**2. `cleanup` now follows the spec shutdown sequence** (`basic/lifecycle.mdx` § Shutdown / stdio):

> the client **SHOULD** initiate shutdown by:
> 1. First, closing the input stream to the child process (the server)
> 2. Waiting for the server to exit, or sending `SIGTERM` if the server does not exit within a reasonable time
> 3. Sending `SIGKILL` if the server does not exit within a reasonable time after `SIGTERM`

Previously `cleanup` sent `SIGTERM` immediately after closing stdin (a well-behaved server never got the chance to exit on EOF) and never escalated, so a server that ignores `SIGTERM` leaked as a running orphan. It now closes stdin, waits `SHUTDOWN_GRACE_PERIOD` (2s), sends `SIGTERM` if still alive, waits again, then sends `SIGKILL` and reaps. Signalling tolerates already-dead processes (`Errno::ESRCH`/`Errno::EPERM`).

**3. Reader-thread robustness:** `handle_line` rescued only `JSON::ParserError`, so a JSON-parseable line that is not an object (e.g. `[]` or `123`) raised `TypeError` and permanently killed the reader thread. Non-object values are now logged and skipped, and the rescue also covers `EncodingError` so a stray undecodable line drops one line instead of the whole transport.

## Tests

New `spec/lib/mcp_client/server_stdio_compliance_spec.rb` (10 examples, written red-first against the old behavior):

- pipes are pinned to UTF-8 under a simulated non-UTF-8 locale (`Encoding.default_external = US-ASCII`), and a real `ruby -e` subprocess response containing `żółć` is delivered instead of deadening the transport
- a real subprocess that exits on stdin EOF is never signalled; a real subprocess that traps `SIGTERM` is escalated to `SIGKILL` and verified dead via `Process.kill(0, pid)`
- mocked-process ordering test for close-stdin → wait → `SIGTERM` → wait → `SIGKILL`, no-signal-on-graceful-exit, and `Errno::ESRCH` tolerance
- `handle_line` skips non-object JSON lines, keeps processing subsequent messages, and survives `EncodingError`

Two existing specs that encoded the old behavior (`#connect` pipe doubles, `#cleanup` immediate-`SIGTERM` expectation) were updated with spec citations.

Full suite: 1330 examples, 0 failures. RuboCop: 111 files inspected, no offenses.

### DIFF
--- a/lib/mcp_client/server_stdio.rb
+++ b/lib/mcp_client/server_stdio.rb
@@ -21,6 +21,12 @@ module MCPClient
     # Timeout in seconds for responses
     READ_TIMEOUT = 15
 
+    # Grace period in seconds allowed at each stage of the shutdown sequence
+    # (after closing stdin, then after SIGTERM) before escalating further, per
+    # MCP 2025-11-25 basic/lifecycle.mdx (Shutdown / stdio): close stdin, wait
+    # for the server to exit, send SIGTERM, then SIGKILL if it still runs.
+    SHUTDOWN_GRACE_PERIOD = 2
+
     # Chunk size (bytes) used when draining the subprocess stderr pipe
     STDERR_READ_CHUNK_SIZE = 8192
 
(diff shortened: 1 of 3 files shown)


simonx1:
External verification (Codex) reviewed this diff: verdict correct with no findings; test coverage (real subprocesses for locale, graceful exit, and SIGKILL escalation) judged adequate.

PR #172, opened 2026-07-20
Fix legacy SSE transport compliance: error responses, endpoint resolution, protocol-version header, init validation

simonx1:

### 1. Deliver JSON-RPC error responses over SSE to the waiting caller

> Implementations **SHOULD** be prepared to handle these error cases: - Protocol version mismatch - Failure to negotiate required capabilities - Request [timeouts]

(MCP 2025-11-25 `basic/lifecycle.mdx`, "Error Handling")

The SSE parser diverted ANY payload with an `error` key to a log/auth path *before* the id check, so every id-bearing JSON-RPC error response from the server was swallowed: the pending request blocked for `read_timeout` and then surfaced as a misleading `Timeout waiting for SSE result` `ToolCallError`, losing the server's error code/message entirely (the error branch in `process_response?` was dead code). Error responses with an id are now stored for the waiter and raised as `MCPClient::Errors::ServerError` carrying the server's code and message, mirroring `JsonRpcCommon#process_jsonrpc_response` on the other transports. Id-less error payloads keep the existing connection-level authorization handling.

### 2. Resolve the `endpoint` event URI against the SSE connection URL

> When a client connects, the server **MUST** send an `endpoint` event containing a URI for the client to use for sending messages. All subsequent client messages **MUST** be sent as HTTP POST requests to this endpoint.

(MCP 2024-11-05 `basic/transports`, "HTTP with SSE" — the deprecated transport this class implements, referenced from 2025-11-25 `basic/transports.mdx` "Backwards Compatibility")

The endpoint data is a URI reference and must be resolved against the SSE connection URL (RFC 3986 §5.1.3, as the reference TypeScript SDK does with `new URL(data, sseUrl)`). Previously the raw value was stored and POSTed against `scheme://host:port` with the SSE URL's path stripped, so a relative endpoint such as `messages?sessionId=x` from a server mounted at `/api/v1/sse` was POSTed to `/messages?sessionId=x` — a URL the server never designated. `handle_endpoint_event` now resolves via `URI.join(@base_url, data)`.

### 3. Send `MCP-Protocol-Version` on all post-initialize HTTP requests

> If using HTTP, the client **MUST** include the `MCP-Protocol-Version: <protocol-version>` HTTP header on all subsequent requests to the MCP server.

(MCP 2025-11-25 `basic/lifecycle.mdx`, "Version Negotiation"; restated in `basic/transports.mdx`, "Protocol Version Header")

`ServerSSE` never captured the negotiated version and never sent the header, so servers keying on it would assume `2025-03-26` and silently downgrade session semantics. `perform_initialize` now captures `result['protocolVersion']` into `@protocol_version`, and both `send_http_request` (all requests/notifications) and `post_jsonrpc_response` (responses to server-initiated requests) add `Mcp-Protocol-Version` whenever it is set — which naturally excludes the initialize POST itself. Header casing matches `HttpTransportBase`.

### 4. Fail the handshake on a non-object initialize result

> After successful initialization, the client **MUST** send an `initialized` notification to indicate it is ready to begin normal operations

(MCP 2025-11-25 `basic/lifecycle.mdx`, "Initialization")

On a truthy non-object initialize result (e.g. a string), `perform_initialize` hit `return unless result.is_a?(Hash)`, skipping both the capability capture and the mandatory `notifications/initialized` — yet `ensure_initialized` still marked the session initialized and the client continued into the Operation phase on a half-completed handshake. `perform_initialize` now cleans up and raises `MCPClient::Errors::ConnectionError` instead.

## Tests

- New `spec/lib/mcp_client/server_sse_compliance_spec.rb` (13 examples, written TDD-first and confirmed failing against the old code) covering: id-bearing error responses stored and raised as `ServerError` without waiting for the timeout, preserved id-less auth handling, relative/path-absolute/absolute endpoint URI resolution plus an end-to-end POST to the resolved endpoint, `@protocol_version` capture, `Mcp-Protocol-Version` presence on `tools/list` / `notifications/initialized` / server-request responses and absence on the initialize POST, and `ConnectionError` (with no `notifications/initialized` and no `@initialized`) on a non-object initialize result.
- Updated four existing tests that encoded the old behavior (raw endpoint storage; id-bearing errors treated as connection-level auth failures/log-only), each with a comment citing the spec basis.
- Full suite: 1333 examples, 0 failures. RuboCop: 111 files inspected, no offenses.

### DIFF
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -407,6 +407,10 @@ module MCPClient
         @connection_established = false
         @sse_connected = false
         @initialized = false # Reset initialization state for reconnection
+        # A fresh session negotiates its own protocol version; keeping the old
+        # one would leak the previous session's version into the next
+        # initialize POST's MCP-Protocol-Version header.
+        @protocol_version = nil
 
         # Reset the SSE parse buffer so a reconnect never inherits a leftover
         # partial event from the previous connection.
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -97,6 +97,9 @@ module MCPClient
       @connection_established = false
       @connection_cv = @mutex.new_cond
       @initialized = false
+      # Negotiated protocol version captured from the initialize result
+      # (sent as the MCP-Protocol-Version header on post-initialize requests)
+      @protocol_version = nil
       @auth_error = nil
       # Whether to use SSE transport; may disable if handshake fails
       @use_sse = true
@@ -685,6 +688,9 @@ module MCPClient
       @rpc_conn.post do |req|
         req.url @rpc_endpoint
         req.headers['Content-Type'] = 'application/json'
+        # MCP lifecycle "Version Negotiation": include the MCP-Protocol-Version
+        # header on all HTTP requests after the initialize handshake.
+        req.headers['Mcp-Protocol-Version'] = @protocol_version if @protocol_version
         @headers.each { |k, v| req.headers[k] = v }
         req.body = json_body
       end
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -101,6 +101,9 @@ module MCPClient
       # (sent as the MCP-Protocol-Version header on post-initialize requests)
       @protocol_version = nil
       @auth_error = nil
+      # Non-auth connection failure cause (e.g. invalid endpoint event URI)
+      # recorded by the SSE worker for wait_for_connection to surface
+      @connection_error = nil
       # Whether to use SSE transport; may disable if handshake fails
       @use_sse = true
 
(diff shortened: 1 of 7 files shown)


simonx1:
External verification (Codex) reviewed this diff. Both actionable findings fixed in a follow-up commit with regression tests: an unresolvable endpoint-event URI now fails the handshake with a TransportError instead of storing the raw value, and cleanup clears the negotiated protocol version so reconnects do not leak the previous session's version header. On the attribution nit: agreed — the MCP-Protocol-Version header requirement originates in the current revision's lifecycle, applied here because this client speaks 2025-11-25 semantics over the deprecated transport.

simonx1:
External delta verification (Codex) re-reviewed the follow-up; the remaining observation is fixed: an unresolvable endpoint URI now records a connection failure that wait_for_connection surfaces immediately as a ConnectionError naming the endpoint, instead of a generic timeout.

PR #173, opened 2026-07-20
Gate optional server features on negotiated capabilities

simonx1:
## What

Implements the Operation-phase MUST from [MCP 2025-11-25 basic/lifecycle](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#operation):

> "Both parties **MUST**: ... Only use capabilities that were successfully negotiated"

Every transport captures `@capabilities` from the initialize result, but the codebase consulted it exactly once (the `tasks.requests.tools/call` gate). The optional server features were requested unconditionally:

- `logging/setLevel` — `Client#log_level=` broadcast to **every** server and failed with a `ServerError` on the first one that never declared `logging`;
- `resources/subscribe` / `resources/unsubscribe` — sent without checking `resources.subscribe`;
- `completion/complete` — sent without checking `completions`;
- `tasks/list` / `tasks/cancel` — sent without checking the server's `tasks.list` / `tasks.cancel` declarations (shapes verified against the official schema.ts `ServerCapabilities`).

## How

- `ServerBase#capability?(*path)` — nil-safe nested lookup over the negotiated capabilities — and `#require_capability!(*path, method:)`, raising the new `MCPClient::Errors::CapabilityError` with a descriptive message instead of sending a request the server never negotiated.
- All four transports gate the four optional-feature methods (the gate runs after connection so it reads the *negotiated* set); `CapabilityError` passes through the existing rescue chains untranslated.
- `Client#log_level=` skips servers without the `logging` capability (debug-logged) instead of failing the whole broadcast; `Client#list_tasks` / `#cancel_task` check `tasks.list` / `tasks.cancel`.

## Tests

New `spec/lib/mcp_client/capability_gating_spec.rb` (11 examples: nested `capability?` semantics, per-method refusal without the capability + no request sent, positive path with the capability, selective `log_level=` broadcast, task-operation gating). 9 failed before the fix. Existing fixtures that exercised these features now declare the corresponding capability (with spec-citation comments). Full suite: 1331 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -605,12 +605,7 @@ module MCPClient
     # @raise [MCPClient::Errors::TaskError] if listing fails
     def list_tasks(cursor: nil, server: nil)
       srv = select_server(server)
-      # Gate only on KNOWN negotiated capabilities: before initialization the
-      # capability set is nil and the request itself triggers the handshake.
-      if capabilities_known?(srv) && !srv.capability?('tasks', 'list')
-        raise MCPClient::Errors::CapabilityError,
-              "Server #{srv.name || srv.class.name} did not declare the tasks.list capability"
-      end
+      ensure_task_capability!(srv, 'list')
 
       params = cursor ? { cursor: cursor } : {}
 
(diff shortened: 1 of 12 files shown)


simonx1:
External verification (Codex) reviewed this diff; both findings fixed in a follow-up commit with regression tests: client-level gates (tasks/list, tasks/cancel, the log_level= broadcast filter) now apply only when the negotiated capability set is known — unconnected servers proceed, and the transport-level gates (which run after the handshake) enforce the lifecycle rule.

simonx1:
External delta verification (Codex) re-reviewed the follow-up; the remaining gap is fixed: task operations on a not-yet-initialized server now trigger the handshake (ping) and re-apply the capability gate against the freshly negotiated set before sending, so a server that negotiates no tasks capability never receives the prohibited request.

PR #174, opened 2026-07-20
Echo related-task _meta on task-related responses; gate taskSupport=required on capability

simonx1:
## What

Two fixes against [MCP 2025-11-25 basic/utilities/tasks](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/tasks) (experimental, but binding on implementations that support tasks — this gem is a task requestor):

### 1. `io.modelcontextprotocol/related-task` echo (MUST)

> "All requests, notifications, and responses related to a task **MUST** include the `io.modelcontextprotocol/related-task` key in their `_meta` field, with the value set to an object with a `taskId` matching the associated task ID."

In the `input_required` flow, a task-augmented `tools/call` leads the server to send `elicitation/create` / `sampling/createMessage` carrying that key — and the client's *responses* are task-related messages that must echo it. The string `related-task` appeared nowhere in `lib/`: handlers never received the incoming `_meta` (so hosts couldn't correlate concurrent tasks), and the Streamable HTTP transport rebuilt responses without `_meta` regardless. Every response the library sent to a task-related request violated the MUST.

Each transport now merges the incoming related-task `_meta` into the outgoing result for elicitation, sampling, and roots responses (shared `ServerBase#merge_related_task_meta`), and the client preserves `_meta` returned by symbol-keyed handler results.

### 2. `taskSupport: "required"` gated on the capability precondition (MUST)

> Rule 1: "If a server's capabilities do not include `tasks.requests.tools.call`, then clients **MUST NOT** attempt to use task augmentation on that server's tools, **regardless of the `execution.taskSupport` value**."

The client applied rule 2.3's 'required' rejection unconditionally, so a tool advertising `taskSupport: "required"` on a server *without* the tasks capability was uncallable through every documented API (`call_tool`, `call_tool_streaming`, and `call_tool_as_task` all raise) — while the spec-correct interaction is a plain `tools/call`. `reject_task_required!` now applies only when the server declares the capability.

## Tests

New `spec/lib/mcp_client/tasks_related_meta_spec.rb` (6 examples: echo on stdio/SSE elicitation+sampling, wire-level Streamable HTTP echo, handler `_meta` preservation, plain-call fallback without the capability). All 6 failed before the fix. The pinned task-required test now declares the capability its scenario presumes.

> Stacked on #159 (`fix/elicitation-compliance`) — builds on its corrected Streamable HTTP response paths. Merge #159 first.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -821,7 +821,10 @@ module MCPClient
     # @param tool_name [String] the tool name (for the message)
     # @raise [MCPClient::Errors::ToolCallError] if the tool requires task execution
     def reject_task_required!(tool, tool_name)
-      return unless tool.task_required?
+      # Tasks Tool-Level Negotiation rule 1: without tasks.requests.tools.call
+      # in the server capabilities, taskSupport is disregarded entirely and
+      # the tool is invoked as a plain call.
+      return unless tool.task_required? && server_supports_task_tool_call?(tool.server)
 
       raise MCPClient::Errors::ToolCallError,
             "Tool '#{tool_name}' requires task-augmented execution; call it with call_tool_as_task instead"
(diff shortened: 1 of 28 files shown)


simonx1:
External verification (Codex) reviewed this diff. On the error-response finding: not applied — the official 2025-11-25 schema gives JSON-RPC error objects no _meta slot (JSONRPCError carries only code/message/data), so a related-task echo on error responses has no schema-valid home; result-shaped responses (including decline/cancel ElicitResults) do echo it. The rule-1 capability precondition was verified fully implemented.

PR #175, opened 2026-07-20
Support OAuth Client ID Metadata Documents (SEP-991)

simonx1:
## What

Implements Client ID Metadata Documents (SEP-991) from the MCP 2025-11-25 authorization spec, which makes CIMD the RECOMMENDED registration mechanism ("**Client ID Metadata Documents**: When client and server have no prior relationship (most common)"):

> Authorization servers and MCP clients **SHOULD** support OAuth Client ID Metadata Documents as specified in [OAuth Client ID Metadata Document]. This approach enables clients to use HTTPS URLs as client identifiers, where the URL points to a JSON document containing client metadata.

> MCP clients **SHOULD** check for this capability and **MAY** fall back to Dynamic Client Registration or pre-registration if unavailable.

Previously the `client_id_metadata_document_supported` AS capability was silently dropped by `ServerMetadata.from_h`, and there was no way to use an HTTPS metadata URL as the `client_id` — against an AS supporting CIMD but not RFC 7591 registration, the flow simply failed (audit finding F29).

Changes:

- **`ServerMetadata`** (`lib/mcp_client/auth.rb`): parses, exposes, and round-trips `client_id_metadata_document_supported` (preserving an explicit `false`), plus a `supports_client_id_metadata_documents?` predicate.
- **`OAuthProvider`** (`lib/mcp_client/auth/oauth_provider.rb`): new `client_id_metadata_url:` option (constructor kwarg + validating writer). Per the spec's client requirements — "The `client_id` URL **MUST** use the \"https\" scheme and contain a path component" — invalid URLs raise `ArgumentError`.
- **`get_or_register_client`** now follows the spec's registration priority order:
  1. Pre-registered/cached client information
  2. Client ID Metadata Documents — when the AS advertises `client_id_metadata_document_supported: true` and a `client_id_metadata_url` is configured, the URL itself becomes the `client_id` (used in authorization and token requests) with no DCR POST
  3. Dynamic Client Registration as a fallback

Serving the metadata JSON document at that URL remains the host application's web-infrastructure duty and is intentionally out of scope.

## Tests

New `spec/lib/mcp_client/oauth_cimd_spec.rb` (20 examples, written red-first): `ServerMetadata` parsing/round-trip of the capability field (including explicit `false`), `client_id_metadata_url` validation (HTTPS scheme, path component, writer assignment, nil clearing), the registration priority order (cached beats CIMD; CIMD beats DCR; DCR fallback when the AS lacks the capability or no URL is configured; `ConnectionError` when nothing is available), persistence of the CIMD `ClientInfo` for the code exchange, and an end-to-end `start_authorization_flow` asserting the metadata URL appears as `client_id` in the authorization URL.

Full suite: 1355 examples, 0 failures. RuboCop: 112 files inspected, no offenses.

## Notes

Stacked on #163 (`fix/oauth-challenge-handling`) and should merge after it.

### DIFF
--- a/lib/mcp_client/auth.rb
+++ b/lib/mcp_client/auth.rb
@@ -226,7 +226,7 @@ module MCPClient
     class ServerMetadata
       attr_reader :issuer, :authorization_endpoint, :token_endpoint, :registration_endpoint,
                   :scopes_supported, :response_types_supported, :grant_types_supported,
-                  :code_challenge_methods_supported
+                  :code_challenge_methods_supported, :client_id_metadata_document_supported
 
       # @param issuer [String] Issuer identifier URL
       # @param authorization_endpoint [String] Authorization endpoint URL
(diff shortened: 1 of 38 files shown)


simonx1:
External verification (Codex) reviewed this diff; both findings fixed in a follow-up commit with regression tests: the CIMD URL validation now requires an https host and path and rejects userinfo, fragments, and dot segments per the Client ID Metadata Document draft; and the OAuth convenience constructors forward client_id_metadata_url.

PR #177, opened 2026-07-20
Support sampling tool calling (SEP-1577) and correct sampling error codes

simonx1:
## What

Implements MCP 2025-11-25 tool calling in sampling (SEP-1577) and corrects the sampling error-code semantics.

`client/sampling.mdx` § Tools in Sampling:

> Clients **MUST** declare support for tool use via the `sampling.tools` capability to receive tool-enabled sampling requests. Servers **MUST NOT** send tool-enabled sampling requests to Clients that have not declared support for tool use via the `sampling.tools` capability.

`schema.ts` `CreateMessageRequestParams.tools` / `.toolChoice`:

> The client MUST return an error if this field is provided but ClientCapabilities.sampling.tools is not declared.

`client/sampling.mdx` § Error Handling:

> Clients **SHOULD** return errors for common failure cases:
> - User rejected sampling request: `-1`
> - Tool result missing in request: `-32602` (Invalid params)
> - Tool results mixed with other content: `-32602` (Invalid params)

Previously `Client#handle_sampling_request` extracted only `messages`/`modelPreferences`/`systemPrompt`/`maxTokens` (plus a fixed extras hash) and **silently dropped `tools` and `toolChoice`** — the host handler sampled without tools while the server believed its tools were being honored — and there was no way to declare `sampling.tools`. Additionally (F42, sampling half), internal handler exceptions and the no-handler fallback both used error code `-1`, which the spec reserves for "User rejected sampling request".

Changes:

- `Client.new(..., sampling_supports_tools: true)` opts the host handler into tool use: the declared capability becomes `"sampling": { "tools": {} }` (plumbed via a new `JsonRpcCommon#declare_sampling_tools` that the handler-derived capability derivation from #160 reads); without the opt-in it stays `"sampling": {}`.
- Arity-5+ sampling handlers now receive the complete `sampling/createMessage` params hash as their fifth argument — `includeContext`/`temperature`/`stopSequences`/`metadata` read exactly as before, plus `tools`, `toolChoice`, `_meta`, and any future fields. Arity 0-4 handler signatures are unchanged (backward compatible).
- A request carrying `tools`/`toolChoice` when `sampling.tools` was not declared is rejected with a JSON-RPC `-32602` (Invalid params) error instead of silently sampling without tools; the handler is never invoked.
- Tool-use responses (`content` array of `type: "tool_use"` blocks, `stopReason: "toolUse"`) pass through faithfully; when the handler returns tool_use content without a `stopReason`, it now defaults to `"toolUse"` instead of `"endTurn"`.
- Error codes: handler exceptions → `-32603` (Internal error); no-handler fallback → `-32601` (Method not found); `-1` is kept exclusively for genuine host rejections (handler returns `nil`), matching the elicitation error semantics already established on the base branch.

## Tests

- New `spec/lib/mcp_client/sampling_tool_use_spec.rb` (17 examples, written red-first): capability declaration (`sampling` vs `sampling.tools`, transport- and Client-level), `-32602` rejection of `tools`/`toolChoice` without the capability, forwarding of the complete params hash (tools/toolChoice/_meta) to arity-5+ handlers, faithful ToolUseContent passthrough with `stopReason: "toolUse"` (including the missing-stopReason default), and the F42 error codes (`-32603` internal / `-32601` no handler / `-1` reserved for rejection).
- Existing `client_sampling_spec.rb` error contexts updated with explicit code assertions and spec-citation comments.
- Full suite: 1361 examples, 0 failures. RuboCop: 113 files inspected, no offenses.

**Stacked on #160** (`fix/handler-derived-capabilities`) — the capability declaration builds on its handler-derived `JsonRpcCommon#client_capabilities`; merge after it.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -1215,21 +1215,26 @@ module MCPClient
     end
 
     # Number of the five positional sampling arguments the handler can accept.
-    # Ruby reports negative arity for optional or rest parameters, and
-    # normalizing that to the minimum required count would starve e.g.
-    # ->(messages, prefs, system, max_tokens, extra = nil) of the raw params
-    # (including the SEP-1577 tools/toolChoice fields), so variable-arity
-    # handlers are sized from Proc#parameters instead: each :req/:opt
-    # parameter accepts one argument and a :rest accepts the full list.
+    # Arity alone cannot size variable-arity handlers: lambdas with optional
+    # or rest parameters report a negative arity, and non-lambda procs with
+    # optional parameters report their mandatory minimum as a nonnegative
+    # arity (proc { |m, p = nil, s = nil, t = nil, extra = nil| }.arity == 1).
+    # Normalizing either to the minimum required count would starve the
+    # handler of the raw params (including the SEP-1577 tools/toolChoice
+    # fields), so any handler whose parameters include :opt or :rest entries
+    # (or whose arity is negative) is sized from Proc#parameters instead:
+    # each :req/:opt parameter accepts one argument and a :rest accepts the
+    # full list. Plain fixed-arity handlers keep arity-based sizing.
     # @return [Integer] how many arguments to pass, capped at 5
     def sampling_handler_arg_count
-      arity = @sampling_handler.arity
-      return [arity, 5].min unless arity.negative?
-
       parameters = @sampling_handler.parameters
       return 5 if parameters.any? { |type, _name| type == :rest }
 
-      [parameters.count { |type, _name| %i[req opt].include?(type) }, 5].min
+      if @sampling_handler.arity.negative? || parameters.any? { |type, _name| type == :opt }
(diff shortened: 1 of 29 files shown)


simonx1:
External verification (Codex) reviewed this diff; the finding is fixed in a follow-up commit with regression tests: handlers with optional or variadic parameters are now sized via Proc#parameters, so a lambda with an optional fifth argument receives the full request params including tools/toolChoice. The -32602/-32603 mappings were verified correct.

simonx1:
External convergence review (Codex) flagged that non-lambda procs with optional parameters were still sized by arity; fixed with regression tests — any handler whose parameters include optional or rest entries is now sized via Proc#parameters, so proc-style handlers with optional trailing arguments receive the full params hash.

simonx1:
Final convergence round (Codex): fix confirmed, no new issues.

PR #178, opened 2026-07-20
Cancel timed-out requests, never re-send them, allow per-request timeouts

simonx1:
## What

Three fixes against [MCP 2025-11-25 basic/lifecycle — Timeouts](https://modelcontextprotocol.io/specification/2025-11-25/basic/lifecycle#timeouts) and the [cancellation utility](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/cancellation):

### 1. Cancellation notification on timeout (SHOULD)

> "When the request has not received a success or error response within the timeout period, the sender **SHOULD** issue a cancellation notification for that request and stop waiting for a response."

`notifications/cancelled` appeared nowhere in `lib/` — on timeout every transport just raised, leaving the server executing work the client had walked away from. All transports now best-effort send `notifications/cancelled` (`requestId` + reason) before raising. Never for `initialize` ("The initialize request **MUST NOT** be cancelled by clients"); task-augmented calls keep using `tasks/cancel`.

### 2. Timed-out requests are no longer silently re-sent

The timeout error was in `with_retry`'s rescue list, so with retries configured a timed-out — possibly still-executing, non-idempotent — `tools/call` was transparently re-sent under a fresh id without cancelling the first attempt. The new `MCPClient::Errors::RequestTimeoutError` (subclass of `TransportError`, existing rescues unaffected) is excluded from the retry loop.

### 3. Per-request timeouts (SHOULD)

> "SDKs and other middleware **SHOULD** allow these timeouts to be configured on a per-request basis."

The only knob was the per-server `read_timeout`, baked into the memoized Faraday connection. `rpc_request(method, params, timeout:)` on every transport and `Client#send_rpc(..., timeout:)` now thread a per-request override to the stdio wait deadline, the SSE result wait, and the individual Faraday request.

### Scope note (honest)

Handling a *received* `notifications/cancelled` for an in-flight server-initiated request (audit finding F47) is recognized in `Client#process_notification` but full stop-processing/suppress-response semantics require moving server-request dispatch off the single read thread on all three transports — a concurrency redesign flagged as follow-up. The spec permits receivers to ignore cancellations they cannot honor.

## Tests

New `spec/lib/mcp_client/timeout_cancellation_spec.rb` (10 examples across stdio/HTTP/SSE: cancellation sent with the abandoned requestId, exactly one send despite retries, no cancellation for initialize, per-request timeout threading, Faraday timeout mapping, client-level forwarding and notification recognition). 9 failed before the fix. Full suite: 1330 examples, 0 failures; RuboCop clean.


simonx1:
External verification (Codex) reviewed this diff; fixed in a follow-up commit with regression tests: the timeout rescue no longer sends notifications/cancelled for task-augmented requests (tasks use tasks/cancel) or for initialize sent through the public rpc paths, and the legacy SSE POST path maps Faraday timeouts to RequestTimeoutError so the no-retry/cancellation semantics apply there too. The SSE parser error-delivery observation is addressed on the #172 branch, which reworks that parser.

PR #179, opened 2026-07-20
Place request _meta at the params level on all transports; add progress tracking

simonx1:
## What

Two related fixes against MCP 2025-11-25 ([`RequestParams._meta`](https://modelcontextprotocol.io/specification/2025-11-25/basic#meta) and [basic/utilities/progress](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/progress)):

### 1. Request-level `_meta` reaches the wire correctly on every transport

> "When a party wants to *receive* progress updates for a request, it includes a `progressToken` in the request **metadata**."

Only `ServerStreamableHTTP` hoisted `_meta` out of the tool arguments into the request params — and only for the symbol `:_meta` key. On stdio, legacy SSE, and plain HTTP, any `_meta` passed through `Client#call_tool`/`get_prompt` was serialized as `params.arguments._meta`: the server saw it as a **tool input argument** (breaking tools with `additionalProperties: false`) and never as request metadata, so progress could not be requested through the documented API on 3 of 4 transports. All transports now share one hoisting helper (both `:_meta` and `'_meta'` accepted), matching what `call_tool_as_task` already did.

### 2. Progress token tracking and per-request routing (SHOULD)

> "Senders and receivers **SHOULD** track active progress tokens"

The library never generated, validated, or tracked tokens — `notifications/progress` fell into the unknown-notification debug branch, reaching only the global `on_notification` firehose with no request association and no cutoff after completion. Now:

```ruby
client.call_tool('longRunning', args, progress: ->(progress, total, message) { ... })
```

auto-generates a unique token, injects it into the request `_meta`, routes matching `notifications/progress` to the callback while the request is active, and unregisters the token on completion so stale notifications are dropped (debug-logged). Raw notifications still reach `on_notification` listeners unchanged.

## Tests

New `spec/lib/mcp_client/progress_meta_spec.rb` (8 examples: `_meta` hoisting on all four transports for tools and prompts including the string-key case, token auto-generation + in-flight routing + stale-token drop, unknown-token tolerance). 7 failed before the fix. Also deduplicates `call_tool`'s inline tool resolution through the existing `resolve_tool`. Full suite: 1328 examples, 0 failures; RuboCop clean.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'logger'
+require 'securerandom'
 
 module MCPClient
   # MCP Client for integrating with the Model Context Protocol
@@ -44,6 +45,9 @@ module MCPClient
         MCPClient::ServerFactory.create(config, logger: @logger)
       end
       @tool_cache = {}
+      # Active progressToken -> callback registrations (MCP progress utility)
+      @progress_callbacks = {}
+      @progress_mutex = Mutex.new
       @prompt_cache = {}
       @resource_cache = {}
       # JSON-RPC notification listeners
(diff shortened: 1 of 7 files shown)


simonx1:
External verification (Codex) reviewed this diff: verdict — fully implements request-level _meta hoisting and progress callbacks across the supported transports; no findings.

PR #188, opened 2026-07-30
Cap gzip response decompression to prevent memory exhaustion

simonx1:
## Summary

Security fix for a medium-severity finding from a codex-security scan (`csf_0c2dfc9f`, `resource-exhaustion.gzip-expansion`).

`ServerStreamableHTTP::JsonRpcTransport#parse_response` decompressed a gzip-encoded response body with a single unbounded `Zlib::GzipReader#read`. Since the peer controls the compression ratio, any connected server could return a small, highly compressed body (a "gzip bomb") that expands to an arbitrarily large string and exhausts host memory before JSON parsing even begins.

## Fix

- Decompress incrementally in 64 KiB chunks into a size-counted buffer.
- Raise `MCPClient::Errors::TransportError` as soon as the expanded output exceeds `MAX_DECOMPRESSED_BODY_BYTES` (64 MiB), instead of materializing the full body.

## Testing

- New spec: a body that expands beyond the (stubbed) limit raises `TransportError` with an "expanded beyond" message.
- Full suite: 1607 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/errors.rb
+++ b/lib/mcp_client/errors.rb
@@ -77,6 +77,14 @@ module MCPClient
     # sender SHOULD cancel and stop waiting, not re-send).
     class RequestTimeoutError < TransportError; end
 
+    # Raised when a response body exceeded the configured size limit (e.g. a
+    # gzip payload that expands past the decompression ceiling). A subclass of
+    # TransportError so existing rescues keep working, but deliberately
+    # excluded from automatic retries: the server already received and
+    # processed the request, so re-sending it could run a non-idempotent
+    # operation again — and would decompress the oversized body each time.
+    class ResponseTooLargeError < TransportError; end
+
     # Raised when tool parameters fail validation against the tool's input
     # schema, or (in strict mode) when a tool result's structuredContent fails
     # validation against the tool's output schema
--- a/lib/mcp_client/json_rpc_common.rb
+++ b/lib/mcp_client/json_rpc_common.rb
@@ -25,7 +25,11 @@ module MCPClient
              Errno::ETIMEDOUT, Errno::ECONNRESET, Errno::EPIPE => e
         # A timed-out request may still be executing server-side; re-sending
         # it could run a non-idempotent operation twice. Never retry those.
+        # An oversized response is the same story from the other direction:
+        # the server already ran the request, so a re-send risks a duplicate
+        # side effect (and re-does the oversized decode).
         raise if e.is_a?(MCPClient::Errors::RequestTimeoutError)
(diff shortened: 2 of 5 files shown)


simonx1:

I asked Codex (`codex-cli 0.146.0`, adversarial review, branch diff vs `main`) to independently answer two questions: was the reported vulnerability real, and does this diff actually fix it. It ran its own reproductions rather than reading only.

**Verdict: `needs-attention`** — the vulnerability is confirmed real, but Codex found two problems in the fix. I reproduced both myself before acting, and both are now fixed in this PR.

### Was the finding real? Yes

> The main-branch vulnerability is real: the client advertises gzip and performs an unbounded `GzipReader#read`; a 16,328-byte reproduction expanded to 16 MiB.

### Finding 1 — `[high]` Oversized responses retried, replaying an already-executed RPC

> The limit is detected only after the peer has received the request and returned a response, yet line 70 raises `TransportError`. `rpc_request` wraps request creation and sending in `with_retry` (`lib/mcp_client/http_transport_base.rb:36-47`), which retries every `TransportError` up to three times by default (`lib/mcp_client/json_rpc_common.rb:24-35`). An oversized response to a non-idempotent `tools/call` is therefore POSTed four times with fresh JSON-RPC IDs. A direct reproduction produced IDs `[1,2,3,4]`. […] The added test calls `parse_response` directly, so it misses this behavior.

**This was correct, and it was my bug.** I reproduced it independently against this branch and got the same result — 4 POSTs, ids `[1,2,3,4]`. The original spec exercised `parse_response` in isolation, which is exactly why it didn't catch it.

Fixed by adding `MCPClient::Errors::ResponseTooLargeError < TransportError` and excluding it from `with_retry`, mirroring the existing `RequestTimeoutError` precedent (same rationale: the server already ran the request, so re-sending risks a duplicate side effect). Same reproduction now reports **1 POST**. Added an `rpc_request`-level regression test asserting exactly one POST.

### Finding 2 — `[medium]` The fixed 64 MiB ceiling was an undocumented compatibility break

> A legitimate 49 MiB binary resource expands past 64 MiB once base64 and JSON framing are included, so the gzipped form is rejected even though the same uncompressed response remains accepted and main had no such limit. This makes response validity depend on the server's encoding choice.

Fair, and the asymmetry is the convincing part — whether a response is accepted shouldn't depend on whether the server chose to gzip it. The ceiling is now the default of a validated per-server `max_decompressed_body_bytes` option, with tests for a raised limit accepting a large legitimate body and for rejecting a non-positive limit.

### Verification after the changes

Full suite 1610 examples / 0 failures, RuboCop clean, and the standalone replay reproduction goes from 4 POSTs to 1.

PR #189, opened 2026-07-30
Reject cross-origin SSE endpoint events

simonx1:
## Summary

Security fix for two medium-severity findings from a codex-security scan (`csf_34259fcd`, `csf_3654a414`, `sensitive-data-exposure.cross-origin-sse-endpoint`).

The legacy SSE transport accepted any absolute URI in the server's `endpoint` control event (`SseParser#resolve_endpoint_uri` did a plain `URI.join` with no origin check). A malicious or compromised peer could therefore point `@rpc_endpoint` at an arbitrary origin, after which:

- every client JSON-RPC POST (`send_http_request`) forwarded the caller's configured headers — `Authorization`, cookies, API keys — to the server-selected origin, and
- every roots/sampling/elicitation callback response (`post_jsonrpc_response`) sent both the configured headers and host-produced callback data there too,

including to private-network hosts reachable from the embedding process.

## Fix

- The endpoint reference is still resolved against the SSE connection URL per RFC 3986 (relative URIs keep working).
- If the resolved endpoint's scheme, host, or port differs from the connection origin, the handshake fails with `MCPClient::Errors::TransportError`, and the failure cause is recorded for the blocked `connect` caller (same pattern as the existing unresolvable-URI path).

This matches the origin pinning other MCP SDKs apply to the legacy HTTP+SSE endpoint event.

## Testing

- New specs: same-origin absolute URIs accepted; cross-origin host, port-only, and scheme-downgrade changes all rejected with the handshake failed.

### DIFF
--- a/lib/mcp_client/server_sse/json_rpc_transport.rb
+++ b/lib/mcp_client/server_sse/json_rpc_transport.rb
@@ -2,10 +2,13 @@
 
 require_relative '../json_rpc_common'
 
+require_relative 'origin_policy'
+
 module MCPClient
   class ServerSSE
     # JSON-RPC request/notification plumbing for SSE transport
     module JsonRpcTransport
+      include OriginPolicy
       include JsonRpcCommon
 
       # Generic JSON-RPC request: send method with params and return result
@@ -178,7 +181,7 @@ module MCPClient
       def create_json_rpc_connection(base_url)
         Faraday.new(url: base_url) do |f|
           f.request :retry, max: @max_retries, interval: @retry_backoff, backoff_factor: 2
-          f.response :follow_redirects, limit: 3
+          f.response :follow_redirects, limit: 3, callback: method(:reject_cross_origin_redirect!)
           f.options.open_timeout = @read_timeout
           f.options.timeout = @read_timeout
           f.adapter Faraday.default_adapter
--- a/lib/mcp_client/server_sse/origin_policy.rb
+++ b/lib/mcp_client/server_sse/origin_policy.rb
@@ -1,0 +1,57 @@
+# frozen_string_literal: true
+
+require 'uri'
+
+module MCPClient
+  class ServerSSE
+    # Origin pinning for the legacy HTTP+SSE transport.
+    #
+    # Everything this transport sends carries the caller's configured headers
+    # (Authorization, API keys, cookies), and callback responses carry
+    # roots/sampling/elicitation data, so no request may leave the origin the
+    # caller connected to — neither by a server-chosen endpoint URI nor by a
+    # redirect.
+    module OriginPolicy
+      # @param base [URI::Generic] the SSE connection URL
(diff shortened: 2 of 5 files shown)


simonx1:

I asked Codex (`codex-cli 0.146.0`, adversarial review, branch diff vs `main`) to independently answer two questions: was the reported vulnerability real, and does this diff actually fix it.

**Verdict: `needs-attention`** — finding confirmed real, but the fix was incomplete. Codex found a bypass; I reproduced it and it is now closed in this PR.

### Was the finding real? Yes

> The vulnerability on main is real — `sse_parser.rb:163` accepted arbitrary absolute targets, while `json_rpc_transport.rb:194-206` and `server_sse.rb:710-717` POST bodies with configured headers.

### Finding — `[high]` Cross-origin 307/308 redirects bypass the endpoint-origin check

> Validation stops after confirming the advertised URI itself is same-origin. JSON-RPC connections still enable redirects at `json_rpc_transport.rb:181`, so the peer can advertise a same-origin endpoint that returns a 307/308 `Location` to another origin. `faraday-follow_redirects` then replays the POST body; it removes only the literal `Authorization` header, leaving configured API-key/custom headers. This affects ordinary RPCs and callback responses because `server_sse.rb:710-717` uses the same connection. I reproduced the replay with the locked Faraday 2.14.3/faraday-follow_redirects 0.5.0: `X-API-Key` and the JSON body reached the foreign origin.

**Correct, and a genuine hole in my fix.** I reproduced it independently against this branch:

```
endpoint accepted: https://example.com/messages     # same-origin, passes my check
RESULT: foreign origin received request
  X-API-Key:     secret-key-123
  Authorization: "ABSENT"                            # middleware strips only this one
  body:          {"jsonrpc":"2.0","id":1,"method":"tools/call",...
```

So origin-pinning the `endpoint` event alone just moves the attack one hop: the peer advertises a legitimate same-origin endpoint and redirects the POST instead.

**Fix:** both the RPC connection and the SSE connection now pass a `callback:` to `follow_redirects` that refuses any hop leaving the SSE origin. It raises `ConnectionError` rather than `TransportError` deliberately — the server has already received the original POST, so `with_retry` must not re-send it (the same reasoning as the retry issue Codex found on #188).

Same reproduction after the fix:

```
raised: MCPClient::Errors::ConnectionError
RESULT: nothing reached the foreign origin
```

While doing this I also noticed the origin helpers were about to exist in three copies, so `same_origin?`/`origin_of`/the redirect guard now live in one `ServerSSE::OriginPolicy` module included by `SseParser`, `JsonRpcTransport` and `ReconnectMonitor`.

### Verification after the changes

New specs cover both directions — a cross-origin redirect is refused with nothing reaching the foreign origin, and a **same-origin** redirect is still followed normally. Full suite 1612 examples / 0 failures, RuboCop clean.

PR #190, opened 2026-07-30
Validate peer-advertised OAuth discovery URLs before fetching (SSRF guard)

simonx1:
## Summary

Security fix for two medium-severity findings from a codex-security scan (`csf_5e895e2db`, `csf_6b8b58c6`, `ssrf.oauth-metadata-discovery`).

Two OAuth discovery inputs are peer-controlled and were fetched with no scheme/origin validation:

1. **Challenge-advertised resource metadata** — `handle_unauthorized_response` extracted the `resource_metadata` URL from a 401 `WWW-Authenticate` header and passed it straight to `@http_client.get`. A remote MCP server could advertise `http://169.254.169.254/...` or any private-network URL and make the host fetch it.
2. **PRM `authorization_servers`** — `discover_via_protected_resource` took the first `authorization_servers` value from the fetched Protected Resource Metadata and issued well-known discovery GETs on that origin. The existing `enforce_https_endpoints!` check only ran *after* these fetches, on endpoints inside the resulting metadata.

## Fix

Apply the existing HTTPS-or-loopback policy (`enforce_https!`, the same one used for discovered AS endpoints) to both values **before** any request is made or state is stored:

- an invalid challenge URL now raises `ConnectionError` and is never stored in `@challenge_metadata_url` (in the transport 401 path this is caught by `process_authorization_challenge`, so the original authorization error still propagates);
- an invalid PRM-advertised authorization server fails discovery before any well-known URL is constructed.

The `http://localhost` / `127.0.0.1` / `[::1]` development exception is preserved, and user-configured server URLs are unaffected — only peer-supplied values are validated.

## Testing

- New specs: non-HTTPS challenge URL rejected without any fetch; non-HTTPS PRM authorization server rejected without any fetch; loopback HTTP allowed for both (dev exception).

### DIFF
--- a/lib/mcp_client/auth/oauth_provider.rb
+++ b/lib/mcp_client/auth/oauth_provider.rb
@@ -194,6 +194,12 @@ module MCPClient
         url = extract_resource_metadata_url(www_authenticate)
         return nil unless url
 
+        # The challenge header is peer-controlled input: enforce the HTTPS
+        # policy on the advertised URL BEFORE it is stored or fetched, so a
+        # malicious challenge cannot pivot this host into plain-HTTP GETs
+        # against internal services (SSRF).
+        enforce_https!(url, 'resource metadata URL (from WWW-Authenticate challenge)')
+
         # Remember the advertised URL even if the fetch below fails, so a
         # later discovery retries it instead of probing well-known URIs the
         # challenge already superseded.
@@ -505,6 +511,12 @@ module MCPClient
           raise MCPClient::Errors::ConnectionError,
                 'Protected resource metadata does not advertise any authorization_servers'
         end
+
+        # authorization_servers is untrusted PRM content: enforce the HTTPS
+        # policy on the advertised origin BEFORE constructing and fetching
+        # well-known URLs on it, so a malicious protected resource cannot
+        # drive discovery GETs against internal services (SSRF).
+        enforce_https!(auth_server_url, 'authorization server (advertised by protected resource metadata)')
 
         server_metadata = fetch_first_server_metadata(authorization_server_metadata_urls(auth_server_url))
         unless server_metadata
--- a/lib/mcp_client/auth/oauth_provider.rb
+++ b/lib/mcp_client/auth/oauth_provider.rb
@@ -68,6 +68,8 @@ module MCPClient
         # separately so a failed fetch is retried authoritatively by discovery.
         @challenge_resource_metadata = nil
         @challenge_metadata_url = nil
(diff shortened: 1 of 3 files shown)


simonx1:

**Verdict: `needs-attention`** — findings confirmed real, but the fix left the sharpest half of the SSRF open. Both issues fixed.

### Finding 1 — `[high]` Peer-controlled loopback URLs still permitted SSRF

> Both new checks reuse `enforce_https!`, whose loopback exception accepts `http://localhost`, `http://127.0.0.1`, and `http://[::1]`. A remote or compromised MCP server can therefore advertise an arbitrary loopback `resource_metadata` path that is fetched at line 210; the same exception applies to the PRM `authorization_servers` check at line 519. This preserves the reported SSRF against **especially sensitive local services**, and the added tests explicitly bless it.

**This is correct and it's the most useful thing the review turned up.** I reused a policy written for *operator-configured* URLs on *peer-supplied* ones. Loopback is not an incidental gap here — it is the highest-value SSRF target (admin panels, metadata services, unauthenticated local daemons), so the fix was weakest exactly where it mattered. And the "added tests explicitly bless it" sting is fair: I wrote a test asserting the hole was fine.

**Fix:** peer-advertised URLs now go through `validate_peer_advertised_url!` — HTTPS required, and loopback/private/link-local hosts refused. The dev exception survives only when the *configured* MCP server is itself local, so pointing at a local stack still works while a remote server can no longer aim us at localhost. The two tests that blessed the old behaviour are replaced by tests asserting rejection, plus one covering the legitimate local-server case.

**Not done:** DNS-resolution checking. A public hostname that resolves to a private address still passes, because that needs resolution-time enforcement in the HTTP layer rather than a string check — I've noted the limitation in the code rather than implying coverage I don't have.

### Finding 2 — `[medium]` Rejected challenges failed open to stale authorization metadata

> The guard raises before setting `@challenge_metadata_url`, but `@challenge_scope` **was already mutated**. Transport challenge processing swallows the validation error, so subsequent discovery sees no pending challenge and may reuse cached AS metadata […] Reproduction showed an invalid HTTP challenge retaining `scope="admin"` and immediately returning the stale cached issuer.

Also correct — I validated in the middle of a multi-field state update. A rejected challenge now clears *all* of its state and records why; discovery re-raises that instead of silently falling back to the cached authorization server the challenge was telling us to replace. Two tests cover it: no scope retained after rejection, and discovery failing closed afterwards.

### Verification

Full suite 1615 examples / 0 failures, RuboCop clean.

PR #191, opened 2026-07-30
Cap the legacy SSE parse buffer to prevent memory exhaustion

simonx1:
## Summary

Security fix for a medium-severity finding from a codex-security scan (`csf_9b731a9e`, `resource-exhaustion.unterminated-sse-buffer`).

`ServerSSE#extract_complete_events` appended every received chunk to `@buffer` and only removed bytes after finding a blank-line SSE terminator. There was no maximum event size, so after the ordinary endpoint handshake a hostile peer could continuously send an unterminated event (refreshing activity with `event:` text) and retain attacker-controlled bytes until host memory was exhausted.

## Fix

- New `MAX_SSE_BUFFER_BYTES` (32 MiB) bounds the *unprocessed partial-event* bytes — complete events inside an oversized chunk are still extracted first, so legitimate large responses are unaffected.
- On overflow the buffer is dropped and the connection fails with `ConnectionError` (same propagation path as the existing embedded-auth-error raise in `process_sse_chunk`), handing control to the existing reconnect logic. Memory stays bounded even against a peer that reconnects and retries forever, since `cleanup` also resets the buffer.

## Testing

- New specs: an unterminated event beyond the (stubbed) cap raises and empties the buffer; a *terminated* event larger than the cap in a single chunk still parses fine.
- Full suite: 1608 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -99,7 +99,9 @@ module MCPClient
       @request_id = 0
       @sse_results = {}
       @mutex = Monitor.new
-      @buffer = ''
+      @buffer = +''
+      # How much of @buffer has already been searched for an event terminator
+      @buffer_scanned = 0
       @sse_connected = false
       @connection_established = false
       @connection_cv = @mutex.new_cond
@@ -430,7 +432,8 @@ module MCPClient
 
         # Reset the SSE parse buffer so a reconnect never inherits a leftover
         # partial event from the previous connection.
-        @buffer = ''
+        @buffer = +''
+        @buffer_scanned = 0
 
         # Log cleanup for debugging
         @logger.debug('Cleaning up SSE connection')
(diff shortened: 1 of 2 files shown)


simonx1:

**Verdict: `needs-attention`** — finding confirmed real, cap confirmed effective for memory, two gaps raised. Both fixed.

### Was the finding real? Yes

> The vulnerability on main was real: `lib/mcp_client/server_sse.rb:899-912` appended peer-controlled chunks indefinitely without a delimiter, while `lib/mcp_client/server_sse/reconnect_monitor.rb:201-206` **disables the SSE response timeout**.

### Finding 1 — `[high]` The capped path still allowed quadratic exhaustion

> The limit is checked only after `@buffer += chunk` and full-buffer delimiter searches. For an unterminated event split across N callbacks, each callback copies and rescans all prior bytes while holding `@mutex`. With ordinary 16 KiB callbacks, reaching 32 MiB requires about 2,049 iterations and **roughly 32 GiB of aggregate string copying** […] Retained memory is bounded, but a hostile peer can still impose severe CPU and allocator pressure before rejection.

**Correct, and it's the more interesting half of the original bug.** I bounded the *memory* and left the *work* unbounded — and the copying happens while holding `@mutex`, so it stalls event processing too. Measured on this branch:

| | 1024 × 16 KiB chunks |
|---|---|
| `+=` and full rescan (before) | **11.2 s** |
| in-place append, incremental scan (after) | **0.018 s** |

The buffer is now appended in place and only newly arrived bytes are scanned, backing up by the longest delimiter minus one so a terminator split across two chunks is still found (there's a test for exactly that). I kept the cap check *after* extraction rather than before appending — checking first looked tempting but wrongly rejects a single oversized chunk that does contain complete events, which an existing test caught.

### Finding 2 — `[medium]` The overflow error bypassed error propagation

> In production this exception originates in Faraday's `on_data` callback and escapes the SSE worker. `handle_sse_connection` resets flags and re-raises from the background thread **without recording `@connection_error`** […] During initial connection callers receive a generic connection failure rather than the size error […] despite the added comment.

Fair — and specifically it contradicted a claim I made in the PR description about handing control to the reconnect path. Overflow now records `@connection_error`, marks the connection down and broadcasts, mirroring the existing endpoint-URI failure path, so `wait_for_connection` surfaces the real reason. Test asserts the recorded cause.

I have **not** rebuilt the worker lifecycle into an explicit terminal-failure/bounded-reconnect policy, which is the rest of that recommendation — that's a broader change to the SSE state machine than this fix warrants, and the reconnect behaviour is now no worse than any other connection failure.

### Verification

Full suite 1611 examples / 0 failures, RuboCop clean.

PR #192, opened 2026-07-30
Cap Streamable HTTP SSE buffers to prevent memory exhaustion

simonx1:
## Summary

Security fix for two medium-severity findings from a codex-security scan (`csf_326917c0`, `csf_9c01b079`, `resource-exhaustion.unterminated-sse-buffer`).

Two SSE parse buffers in `ServerStreamableHTTP` grew without bound while waiting for a blank-line event terminator:

1. **GET events stream** — `process_event_chunk` appended every chunk to `@buffer`, removing bytes only after a complete event was found. A peer could continuously send delimiter-free chunks (each arriving before the socket inactivity timeout) and grow host memory until exhaustion.
2. **Resumption GET** — `issue_resumption_get` accumulated chunks in a local buffer with no size cap and no byte budget inside the blocking request, so a peer could keep the socket active with delimiter-free replay data and exhaust the resumption worker.

## Fix

- New `MAX_SSE_BUFFER_BYTES` (32 MiB) bounds *unprocessed partial-event* bytes in both paths, enforced by a shared `enforce_sse_buffer_cap!` helper. Complete events inside an oversized chunk are still extracted before the check, so legitimate large events are unaffected.
- Events stream: on overflow the buffer is dropped and `ConnectionError` propagates to the events loop, which treats it like any connection failure (reconnect with existing backoff).
- Resumption GET: on overflow the GET aborts (rescued by the existing `StandardError` handler); the deadline-bounded resumption loop decides whether to retry, with a fresh buffer per attempt.

## Testing

- New specs: unterminated events-stream data beyond the (stubbed) cap raises and empties the buffer; terminated oversized events still parse; delimiter-free resumption data trips the cap and aborts the GET without raising out of the worker.

### DIFF
--- a/lib/mcp_client/server_streamable_http.rb
+++ b/lib/mcp_client/server_streamable_http.rb
@@ -33,6 +33,13 @@ module MCPClient
     SSE_MAX_RECONNECT_DELAY = 30   # Maximum reconnect delay in seconds
     THREAD_JOIN_TIMEOUT = 5 # Timeout for thread cleanup
 
+    # Maximum bytes an SSE parse buffer (events stream or resumption GET) may
+    # hold while waiting for an event terminator. The stream is
+    # peer-controlled: without a cap, a hostile server could withhold the
+    # blank-line delimiter forever and grow the buffer until the host runs
+    # out of memory. Generous enough for any legitimate JSON-RPC event.
+    MAX_SSE_BUFFER_BYTES = 32 * 1024 * 1024
+
     # @!attribute [r] base_url
     #   @return [String] The base URL of the MCP server
     # @!attribute [r] endpoint
@@ -803,6 +810,10 @@ module MCPClient
         req.options.on_data = proc do |chunk, _bytes|
           buffer << chunk
           process_resumption_buffer(buffer, state)
+          # A peer replaying delimiter-free data must not grow this buffer
+          # without bound: abort the GET (rescued below); the deadline-bounded
+          # resumption loop decides whether to retry with a fresh buffer.
+          enforce_sse_buffer_cap!(buffer)
         end
       end
     rescue StandardError => e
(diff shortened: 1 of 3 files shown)


simonx1:

**Verdict: `needs-attention`** — findings confirmed real, memory bound confirmed effective, one gap raised and fixed.

> The pre-change vulnerability was real in both buffers, and the diff bounds retained memory […]

### Finding — `[high]` The cap still permitted quadratic CPU and allocation exhaustion

> Each resumption chunk is appended and then `process_resumption_buffer` searches the entire accumulated buffer from byte zero; the cap runs only afterward. With the default Net::HTTP 16 KiB chunks, reaching 32 MiB requires about 2,048 progressively larger scans — roughly 32 GiB of inspected data. The events path at lines 892-905 is worse because `@buffer += chunk` also **copies** the growing buffer before two full delimiter searches. Local reproduction took 6.3 seconds to scan only 16 MiB […] while `retry: 0` makes reconnects immediate, converting the original memory DoS into a sustained CPU/allocator DoS.

**Correct** — and the same critique landed on the legacy SSE buffer in #191, where I measured the identical pattern at **11.2 s → 0.018 s** after fixing it. I bounded the memory and left the work unbounded, which for an attacker is a fine substitute.

The `retry: 0` amplification point is a nice piece of cross-PR reasoning: #193/#200 floor those reconnects, but a peer that can make each connection expensive still gets leverage from repeating them.

**Fix (both paths):**
- `process_event_chunk` appends in place and scans only newly arrived bytes.
- `process_resumption_buffer` keeps a scan cursor in the per-stream `state` hash, so concurrent resumption workers stay independent rather than sharing a cursor.
- Both back up by the longest delimiter minus one, so a terminator split across two chunks is still found — there's a test for that.
- The caps still run after extraction, so a single oversized chunk containing complete events is parsed rather than rejected.

### Verification

New tests assert both paths stay under 1 s for 512 × 16 KiB delimiter-free chunks (the old code needed multiple seconds for the same input), plus the split-terminator case. Full suite 1613 examples / 0 failures, RuboCop clean.

PR #193, opened 2026-07-30
Floor server retry directives on the events stream to prevent reconnect storms

simonx1:
## Summary

Security fix for a medium-severity finding from a codex-security scan (`csf_16ba60d1`, `resource-exhaustion.zero-delay-reconnect`).

`events_reconnect_delay` honored the server's SSE `retry:` directive literally and preferred it over the exponential-backoff fallback. A malicious server could return `retry: 0` and close each events stream immediately, driving a tight reconnect loop — sustained CPU, TLS-handshake, and connection churn with no backoff for as long as the client stayed connected.

## Fix

- New `MIN_EVENTS_RECONNECT_DELAY` (100 ms) floors peer-supplied retry directives on the long-lived GET events stream.
- Directives ≥ 100 ms are honored exactly as before; only zero/near-zero values are floored, bounding worst-case churn at 10 reconnects/second instead of a busy loop.
- Spec compliance: SEP-1699 says the client MUST wait "the given number of milliseconds before attempting to reconnect" — the retry field is a lower bound, so waiting longer remains compliant. The existing spec that asserted a literal `0` delay is updated accordingly (parsing `retry: 0` as *valid* is unchanged).
- The SEP-1699 resumption loop intentionally keeps honoring `retry: 0` for the documented immediate-poll pattern — it is already bounded by the request deadline.

## Testing

- Updated/new specs: `retry: 0` still parses but the events reconnect delay floors to the minimum; a 10 ms directive floors as well; a 250 ms directive is honored exactly (existing spec, unchanged).
- Full suite: 1607 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/server_streamable_http.rb
+++ b/lib/mcp_client/server_streamable_http.rb
@@ -33,6 +33,14 @@ module MCPClient
     SSE_MAX_RECONNECT_DELAY = 30   # Maximum reconnect delay in seconds
     THREAD_JOIN_TIMEOUT = 5 # Timeout for thread cleanup
 
+    # Floor for server-supplied retry directives on the long-lived events
+    # stream. The directive is peer-controlled: honoring "retry: 0" literally
+    # would let a hostile server that closes every stream drive a tight
+    # reconnect loop (sustained CPU/TLS/connection churn). Waiting longer
+    # than the directive stays SEP-1699 compliant — the retry field is a
+    # lower bound on the reconnect delay, not an exact schedule.
+    MIN_EVENTS_RECONNECT_DELAY = 0.1
+
     # @!attribute [r] base_url
     #   @return [String] The base URL of the MCP server
     # @!attribute [r] endpoint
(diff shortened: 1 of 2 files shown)


simonx1:

**Verdict: `needs-attention`** — finding confirmed real, fix confirmed correct for the path it covers, one gap raised. That gap is real but is already fixed by sibling PR #200; details below.

### Was the finding real? Yes, and the spec argument holds

> The reported main-branch vulnerability was real: `lib/mcp_client/server_streamable_http.rb:729-731` converted the truthy integer `0` to a 0.0-second delay, consumed by every long-lived reconnect path at lines 697-718. HEAD lines 740-744 correctly floor that path without changing fallback or larger directives.

It also independently checked the compliance argument in the PR description, which I'd flagged as the one judgement call worth reviewing:

> The 100 ms floor is compliant: the [MCP transport specification](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server) requires waiting for the directive, while the [standard SSE reconnect algorithm](https://html.spec.whatwg.org/multipage/server-sent-events.html#reestablish-the-connection) explicitly permits waiting longer.

### Finding — `[high]` `retry: 0` still spins the *resumption* GET loop

> A server can answer a POST SSE stream with `retry: 0`, an event ID, and no JSON-RPC response; `resume_or_fail` then starts this worker. Lines 787 and 795 preserve a zero delay […] Exercising the actual `parse_sse_response` path with networking stubbed produced **268,252 attempts in 55 ms**, confirming there is no client-side rate bound.
>
> Recommendation: Apply a positive minimum to both resumption-delay assignments […]

**The finding is accurate** — this branch floors only `events_reconnect_delay`, and the resumption loop keeps a literal zero. Codex reviewed each branch against `main` in isolation (deliberately, so it wouldn't assume sibling work), so it flagged this as missing.

It is, however, **already implemented in #200**, which adds `MIN_RESUMPTION_RECONNECT_DELAY = 0.01` and routes both assignments through a `resumption_delay` helper — precisely the recommendation. I've deliberately not duplicated it here: two branches defining the same constant would conflict for no behavioural gain.

**Worth being explicit about the merge implication:** if this PR lands and #200 does not, the resumption path remains exposed to the same hostile directive. They should land together, or #200 first.

I kept the two floors as separate values on purpose — 100 ms for the long-lived events stream, 10 ms for resumption — because SEP-1699's polling pattern legitimately wants fast reconnects there, and 10 ms is enough to prevent a busy loop.

### Verification

No code change on this branch in response to the review. Full suite 1607 examples / 0 failures, RuboCop clean, unchanged from the original push.

<sub>Codex review requested and summarized by Claude Code; the quoted text is Codex's, the assessment is mine.</sub>

PR #194, opened 2026-07-30
Bound server-initiated response POST threads

simonx1:
## Summary

Security fix for two medium-severity findings from a codex-security scan (`csf_2c81df29`, `csf_19a3c0f2`, `resource-exhaustion.unbounded-thread-fanout`).

In `ServerStreamableHTTP`, every server-initiated message needing a reply spawned an untracked `Thread.new` performing a blocking HTTP POST:

- `handle_ping_request` created one thread per remotely supplied ping, and
- `post_jsonrpc_response` created one thread per roots/sampling/elicitation reply and per error response (including `-32601` replies to *unknown* methods — so arbitrary garbage requests also fanned out threads).

No pool, semaphore, or queue bounded the fan-out, so a peer flooding requests on the events stream (or holding reply POSTs open) could accumulate Ruby threads and connections until the host was exhausted.

## Fix

- Pongs now go through `post_jsonrpc_response`, removing the duplicated POST code and second spawn site; wire behavior (headers, body, endpoint) is unchanged.
- That single path is gated by a concurrency budget: `MAX_CONCURRENT_RESPONSE_POSTS` (8). A slot is acquired before spawning and released when the POST finishes; when the budget is exhausted the response is dropped with a warning — backpressure against a flooding peer, ample headroom for legitimate traffic.

## Testing

- New specs: at budget, neither a response nor a pong spawns a thread (and a drop warning is logged); the slot count returns to zero after a POST completes. The existing POST-SSE-stream ping spec still verifies pong wire format end-to-end.
- Full suite: 1609 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/server_streamable_http.rb
+++ b/lib/mcp_client/server_streamable_http.rb
@@ -33,6 +33,14 @@ module MCPClient
     SSE_MAX_RECONNECT_DELAY = 30   # Maximum reconnect delay in seconds
     THREAD_JOIN_TIMEOUT = 5 # Timeout for thread cleanup
 
+    # Ceiling on concurrent threads POSTing server-initiated responses
+    # (pongs, roots/sampling/elicitation replies, error responses). Each
+    # server request on the events stream costs one blocking HTTP POST in
+    # its own thread; without a bound, a peer flooding requests could
+    # accumulate threads and connections until the host is exhausted.
+    # Responses beyond the budget are dropped with a warning.
+    MAX_CONCURRENT_RESPONSE_POSTS = 8
+
     # @!attribute [r] base_url
     #   @return [String] The base URL of the MCP server
     # @!attribute [r] endpoint
@@ -113,6 +121,7 @@ module MCPClient
       @last_event_id = nil
       @sse_retry_ms = nil
       @pending_stream_responses = {}
+      @response_post_count = 0
       @oauth_provider = opts[:oauth_provider]
 
       # SSE events connection state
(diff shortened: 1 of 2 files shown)


simonx1:

**Verdict: `needs-attention`** — findings confirmed real, the cap itself confirmed correct, three problems raised. Two are fixed here; the third is a design question I'd like your call on.

> The main-branch vulnerability is real: both pongs (`server_streamable_http.rb:988-1007` on main) and other server-request replies (`:1186-1208` on main) unconditionally spawned blocking POST threads. The diff atomically limits successfully started workers to eight […]

### Finding 1 — `[medium]` Thread creation failure permanently leaks a slot — **fixed**

> The counter is incremented before `Thread.new`, but the release ensure is inside the new thread's block. If `Thread.new` raises `ThreadError`, no block runs and the reservation is never released. This was reproduced directly: one forced `ThreadError` left `@response_post_count` at 1. After eight transient creation failures, this instance drops every future reply, **including after reconnect**, because cleanup does not reset the counter.

Straightforwardly my bug, and the failure mode is nasty precisely because it is silent and permanent — thread exhaustion is exactly when `Thread.new` starts failing, so the leak triggers under the conditions the cap exists for. The spawn is now isolated so the slot is released if it fails, with a test asserting the counter returns to zero.

### Finding 2 — `[medium]` Unbounded warning-log amplification — **fixed**

> Every rejected request synchronously emits a WARN containing the full peer-controlled request ID. WARN is enabled by the default logger […] so once eight POSTs are stalled an attacker can send unlimited pings and force unlimited output and string allocation without encountering the thread cap.

Also correct, and a good framing: the cap swapped one attacker-controlled resource for another. Saturation is now logged at most once per 5 s, without the peer-supplied id, with a cumulative dropped counter so suppressed events stay visible. Test asserts 20 drops produce exactly one warning containing no peer id.

### Finding 3 — `[high]` Dropping replies is itself a protocol violation — **not fixed; your call**

> At capacity, this returns without posting, queuing, retrying, or failing the session. […] MCP requires prompt ping responses and defines requests as expecting matching responses. Worse, **callbacks execute before admission** at lines 1042, 1061, and 1080, so an expensive sampling or user interaction can complete before its result is discarded; a server retry can repeat that work.

I think this is right as a critique, and the callback-ordering point is the sharpest part — we can prompt a user, get their answer, then throw it away. But I don't think it should be fixed inside this PR:

- The fix Codex proposes (bounded worker pool + queue with backpressure, reserve capacity *before* invoking callbacks, or fail the session on overload) is a redesign of the server-request path, not a bound on it.
- The choice here was between *unbounded threads* (the reported vulnerability) and *bounded threads with drops*. Dropping is a strictly smaller problem than exhausting the host, and 8 concurrent in-flight replies is far above any legitimate traffic — a well-behaved server never reaches it.

So I'd suggest landing this as the containment fix and tracking the queue/backpressure redesign separately. Happy to do it here instead if you'd prefer it in one go.

### Verification

Full suite 1611 examples / 0 failures, RuboCop clean.

PR #195, opened 2026-07-30
Discard unsolicited SSE responses and clear results across reconnects

simonx1:
## Summary

Security fix for a medium-severity finding from a codex-security scan (`csf_c1c82c45`, `resource-exhaustion.unsolicited-response-state`).

In the legacy SSE transport, `SseParser#process_response?` stored **every** id-bearing server message into `@sse_results` — with no check that the id belonged to an outstanding client request and no size limit. `check_for_result` deleted only the exact id a caller was waiting for, and `cleanup` reset the parse buffer but never cleared the map. A malicious server could therefore stream valid SSE response events with fresh attacker-chosen ids (or late responses to timed-out requests) and grow client memory without bound, persisting across reconnects for the lifetime of the client.

## Fix

- New `@pending_request_ids` set tracks ids a caller is actively waiting on. `send_jsonrpc_request` registers the id **before** posting (the SSE stream may deliver the response before the POST returns) and unregisters it — dropping any unconsumed result — when the request completes, fails, or times out.
- `process_response?` discards responses whose id is not outstanding, with a debug log.
- `cleanup` clears `@sse_results`, so undelivered results cannot survive reconnects.

Memory for peer-controlled response state is now bounded by the number of concurrently outstanding client requests.

## Testing

- New specs: unsolicited response ids are discarded; the pending id is unregistered after a request completes; `cleanup` empties `@sse_results`.
- Updated specs that pre-seeded responses now register the id as outstanding first (matching production ordering, where the response-provoking POST happens after registration).
- Full suite: 1609 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -429,10 +429,15 @@ module MCPClient
         # partial event from the previous connection.
         @buffer = ''
 
-        # Drop undelivered results: their waiters fail with the connection,
-        # and retaining entries across reconnects would let stale peer data
-        # accumulate for the lifetime of the client.
-        @sse_results = {}
+        # Drop results nobody is waiting for, so peer-supplied state cannot
+        # accumulate across reconnects. Results for still-pending requests are
+        # KEPT: a response can arrive while its POST is still returning, and
+        # the waiter (which reconnects through ensure_sse_connection_active)
+        # is about to consume it. Discarding those reported a timeout for a
+        # tool call the server had already executed — inviting a duplicate
+        # manual retry. unregister_pending_request clears each entry when its
+        # request finishes, so nothing lingers.
+        @sse_results.select! { |id, _| @pending_request_ids.include?(id) }
 
         # Log cleanup for debugging
         @logger.debug('Cleaning up SSE connection')
--- a/lib/mcp_client/server_sse.rb
+++ b/lib/mcp_client/server_sse.rb
@@ -91,6 +91,10 @@ module MCPClient
       @tools_data = nil
       @request_id = 0
       @sse_results = {}
+      # Ids of requests a caller is actively waiting on. Only responses for
+      # these ids are stored in @sse_results — everything else on the peer
+      # controlled stream is unsolicited and discarded.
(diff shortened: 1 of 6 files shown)


simonx1:

**Verdict: `needs-attention`** — finding confirmed real, guard confirmed sound, but my `cleanup` change introduced a data-loss race. Fixed.

### Was the finding real? Yes, and the guard works

> The vulnerability on main was real: `lib/mcp_client/server_sse/sse_parser.rb:100-109` unconditionally retained every truthy response id, while `json_rpc_transport.rb:271-275` removed only requested ids and cleanup never cleared the hash. The new pending-id guard at `sse_parser.rb:105-108` bounds unsolicited state […]

### Finding — `[high]` Reconnect cleanup could erase a valid completed response

> A legitimate response can be stored while the POST is returning. If the SSE stream then becomes inactive, `wait_for_sse_result` checks connectivity before consuming the result and invokes `cleanup` […] The unconditional reset here deletes that response. A deterministic reproduction timed out on this branch, while the same flow with main's cleanup returned the successful result. **This can report failure after a non-idempotent tool already executed, encouraging a duplicate manual retry.**

**This was a regression I introduced**, and the consequence is the sharp part: the client reports a timeout for work the server has already done, so a human or a wrapper retries a tool call that already ran. That's the same duplicate-side-effect hazard #196 exists to prevent, reintroduced from a different direction.

The trace holds up against the code: `ensure_sse_connection_active` → `cleanup` → `@sse_results = {}`, and only *then* does `wait_for_result_with_timeout` look for the result.

**Fix:** `cleanup` now keeps results whose ids are still pending and drops everything else:

```ruby
@sse_results.select! { |id, _| @pending_request_ids.include?(id) }
```

The anti-accumulation property is preserved — unsolicited and late responses are still discarded, and `unregister_pending_request` already clears each entry when its request finishes, so nothing lingers after a caller is done.

### Verification

Two regression tests added: a pending id's result survives `cleanup` while an unsolicited one is dropped, and a result delivered through the real SSE event path immediately before a reconnect is still returned to `check_for_result`. Full suite 1611 examples / 0 failures, RuboCop clean.

PR #196, opened 2026-07-30
Never auto-retry non-idempotent tools/call requests

simonx1:
## Summary

Batched security fix for four low-severity findings from the codex-security scan (`csf_11def2fd` SSE, `csf_a4cbaa88` Streamable HTTP, `csf_9f56a0b6` plain HTTP, `csf_a1df8a81` stdio — `request-replay.non-idempotent-retry`).

`with_retry` re-sent any request after failures classified as transient (HTTP 5xx → `TransientServerError`, `TransportError`, `IOError`, connection resets). But an "ambiguous" failure can arrive **after** the server received and started executing the request — so an automatic retry of `tools/call` could run a non-idempotent operation twice. JSON-RPC provides no idempotency key to make the duplicate safe, and `ServerHTTP` enables retries by default (`retries: 3`).

## Fix

- `with_retry` now takes the JSON-RPC method being sent; `rpc_request` passes it on all four transports.
- Methods in `NON_IDEMPOTENT_METHODS` (currently `tools/call`) fail fast on any error instead of being replayed. Everything else (`tools/list`, `resources/read`, `ping`, custom methods, …) retries exactly as before.
- This extends the existing posture (plain `ServerError` and `RequestTimeoutError` were already never retried, for the same reason) to cover the ambiguous-failure classes.

**Behavior change**: a `tools/call` that hits a 5xx or dropped connection now surfaces the error to the caller instead of silently re-POSTing. Two specs that asserted the old replay behavior are updated to use an idempotent method; new specs on each transport pin the fail-fast behavior.

## Testing

- New specs per transport (HTTP-base shared path, SSE, stdio): `tools/call` is attempted exactly once on transient failure; idempotent methods still retry to success.
- Full suite: 1611 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/http_transport_base.rb
+++ b/lib/mcp_client/http_transport_base.rb
@@ -33,7 +33,7 @@ module MCPClient
     def rpc_request(method, params = {}, timeout: nil)
       ensure_connected
 
-      with_retry do
+      with_retry(method) do
         request_id = @mutex.synchronize { @request_id += 1 }
         request = build_jsonrpc_request(method, params, request_id)
         begin
--- a/lib/mcp_client/json_rpc_common.rb
+++ b/lib/mcp_client/json_rpc_common.rb
@@ -3,6 +3,14 @@
 module MCPClient
   # Shared retry/backoff logic for JSON-RPC transports
   module JsonRpcCommon
+    # JSON-RPC methods with arbitrary side effects that MUST NOT be re-sent
+    # automatically. Even a "transient" failure (5xx, dropped connection,
+    # malformed response) can arrive AFTER the server received the request,
+    # so a retry could execute the operation twice — and JSON-RPC has no
+    # idempotency key to make the duplicate safe. Callers who want to retry
+    # such an operation must decide that explicitly.
+    NON_IDEMPOTENT_METHODS = %w[tools/call].freeze
+
     # Execute the block with retry/backoff for transient errors only.
     #
     # Retries genuinely transient failures where the request most likely did not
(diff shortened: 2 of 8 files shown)


simonx1:

**Verdict: `approve` — no material findings.** This was the only security PR in the batch that came back clean on the first pass.

> Ship. The vulnerability was real on main: `lib/mcp_client/json_rpc_common.rb:20-35` retried ambiguous transport/5xx failures around `tools/call`. The new guard at `lib/mcp_client/json_rpc_common.rb:43-45` preserves the original exception and prevents replay. Method propagation covers HTTP and Streamable HTTP via `lib/mcp_client/http_transport_base.rb:36`, SSE at `lib/mcp_client/server_sse/json_rpc_transport.rb:22`, and stdio at `lib/mcp_client/server_stdio/json_rpc_transport.rb:121`. Targeted execution reproduced **three attempts on main versus one after the fix**, while idempotent requests still retry. No material regression found.

Two things worth noting from that:

- It independently confirmed the propagation is complete across **all four** transports, which was the part most likely to be half-done — this fix is only as good as its least-covered call site.
- It verified the behaviour change is scoped: idempotent methods still retry. That's the property that keeps this from being a availability regression, and it's the one I'd want a reviewer to check hardest.

Related: while reviewing sibling PRs, Codex found the same replay hazard reachable through *other* error paths that this PR doesn't cover — an oversized gzip response (#188) and a refused cross-origin redirect (#189), both of which raised retryable `TransportError`s after the server had already executed the request. Both are now non-retryable on their own branches. This PR's method-level guard and those error-level guards are complementary: one covers "which method", the other "which failure".

### Verification

Unchanged since the original push: full suite 1611 examples / 0 failures, RuboCop clean.

<sub>Codex review requested and summarized by Claude Code; the quoted text is Codex's, the framing is mine.</sub>

PR #197, opened 2026-07-30
Bound server-supplied schema pattern matching with a timeout

simonx1:
## Summary

Batched security fix for two low-severity findings from the codex-security scan (`csf_4b941e97`, `csf_6b2310a1`, `redos.schema-pattern-validation`).

Both validators compiled and evaluated a `pattern` keyword taken directly from remote input with no time bound:

- `SchemaValidator.validate_pattern` — pattern comes from the tool's `outputSchema`, applied to every structured tool result.
- `ElicitationValidator.validate_string_value` — pattern comes from the server's elicitation `requestedSchema`.

A pathological expression could therefore monopolize the calling thread.

## Fix

- New `PATTERN_MATCH_TIMEOUT` (1.0s) on both validators; patterns are compiled with `Regexp.new(pattern, timeout:)`.
- A timeout now produces a **validation error**, not a skip. This matters: `Regexp::TimeoutError` descends from `RegexpError`, so the pre-existing "an invalid pattern is not enforced" rescue would have swallowed a timeout and *accepted* a value whose constraint was never evaluated. Genuinely invalid patterns keep the old unenforced behavior.
- The elicitation pattern check moved into its own `validate_string_pattern` helper so both call sites read the same way.

## Scoping note (updated after Codex review)

My original note here hedged that Ruby >= 3.2 memoizes regex backtracking and that I could not construct a payload that hangs on 3.3. That was too weak: memoization is bypassed by **backreferences**, and Codex found a pattern of the shape `\A((a|aa)*)\1+\z` that stayed stuck well beyond 2 seconds on main. The vulnerability is real in the strong sense, not merely defence-in-depth.

Codex also showed that a *per-match* timeout is not an overall bound, since the peer chooses how many strings are matched. The fix now applies a single deadline to the whole validation. See the review comment on this PR for the reproductions.

## Testing

- New specs on both validators: the compiled Regexp carries the timeout, and a real timeout (ambiguous alternation over a 40 KB input with a 1 ms stubbed budget — no mocking of the match itself) produces an error instead of silently accepting the value.

### DIFF
--- a/lib/mcp_client/elicitation_validator.rb
+++ b/lib/mcp_client/elicitation_validator.rb
@@ -18,6 +18,11 @@ module MCPClient
     # Allowed string formats per MCP spec
     STRING_FORMATS = %w[email uri date date-time].freeze
 
+    # Wall-clock budget for matching one server-supplied `pattern`. The
+    # requestedSchema comes from the remote server, so an expensive
+    # expression must not be able to monopolize the calling thread.
+    PATTERN_MATCH_TIMEOUT = 1.0
+
     # Validate that a requestedSchema conforms to MCP elicitation constraints.
     # Returns an array of error messages (empty if valid).
     # @param schema [Hash] the requestedSchema
@@ -183,15 +188,7 @@ module MCPClient
         errors << "Field '#{field}' must be one of: #{allowed.join(', ')}" unless allowed.include?(value)
       end
 
-      if prop['pattern']
-        begin
-          unless value.match?(Regexp.new(prop['pattern']))
-            errors << "Field '#{field}' must match pattern '#{prop['pattern']}'"
-          end
-        rescue RegexpError
-          # Skip pattern validation if the pattern is invalid
-        end
-      end
+      errors.concat(validate_string_pattern(field, value, prop['pattern'])) if prop['pattern']
 
       if prop['minLength'] && value.length < prop['minLength']
         errors << "Field '#{field}' must be at least #{prop['minLength']} characters"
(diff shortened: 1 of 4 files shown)


simonx1:

**Verdict: `needs-attention`** — and it corrected me on two things. Both are now addressed.

### It found the ReDoS case I said I couldn't construct

My PR description hedged that Ruby ≥ 3.2's memoization defeats classic ReDoS payloads and that I couldn't make one hang. Codex found one:

> Main was genuinely vulnerable despite Ruby's memoization: a backreference pattern remained stuck beyond 2 seconds.

Backreferences bypass the memoization optimisation — `\A((a|aa)*)\1+\z` is the shape. So the finding was real on main in the strong sense, not just as defence-in-depth. I've dropped the hedge from the PR description.

### Finding — `[high]` Per-match timeouts multiply into an unbounded total

> Each recursive array item is validated at line 229, while lines 267-271 grant **every pattern match a fresh one-second budget** and then continue after timeout. A server controls both `outputSchema` and `structuredContent`, so it can return many short strings under one pathological `items.pattern`. Using `\A((a|aa)*)\1+\z` and three 51-byte nonmatching strings took 3.00 seconds; N entries cost approximately N seconds of CPU. The elicitation path has the same multiplication […] Thus the claimed protection against monopolizing the caller thread is not an overall bound.

**This is correct and it undercut the fix's actual claim.** Bounding each match while the peer chooses the number of matches isn't a bound at all. My own reproduction with ten items consumed the full ~10s.

**Fix:** both validators now start **one monotonic deadline per top-level call** and hand each match only the remaining budget (floored at 10 ms so a nearly-exhausted budget still makes progress rather than failing everything left). An exhausted budget yields a validation error, preserving the fail-closed semantic — a constraint that could not be evaluated never counts as satisfied.

Same reproduction after the change:

```
10 items, total validation time: 1.00s (per-match limit would allow ~10s)
```

### Verification

New specs assert the aggregate bound directly — 8 items / 8 fields under a 0.3s budget must finish well inside 1.5s, where a per-match limit would allow ~2.4s. Full suite 1612 examples / 0 failures, RuboCop clean.

PR #198, opened 2026-07-30
Keep secrets and payloads out of logs; stop reflecting host errors to peers

simonx1:
## Summary

Batched security fix for **seven** low-severity findings from the codex-security scan, grouped because they all concern data crossing a trust boundary it shouldn't (host → log, or host → remote peer).

### Logging hygiene (4 findings)

| Finding | Problem |
|---|---|
| `csf_8af6f441` | `Client#initialize` logged each server config with `#inspect` — including `Authorization`/API-key headers and the stdio subprocess `env`. |
| `csf_d48a863c` | Every outbound JSON-RPC request was logged in full, so DEBUG recorded `tools/call` arguments verbatim. |
| `csf_e87cee75` | Every HTTP response body was logged in full, recording tool results and resource contents. |
| `csf_1ad7603f` | `notifications/message` content from the server was written verbatim at a peer-chosen severity. |

Fixes:

- **Config redaction** — `SENSITIVE_CONFIG_KEYS` (headers, env, token, api_key, …) are replaced with `[REDACTED]` before logging; nested hashes have every value redacted. Non-sensitive fields still log normally.
- **Payload omission** — new shared `describe_jsonrpc_message` / `describe_body_size` helpers in `JsonRpcCommon`; all four transports now log `tools/call id=7` and `200 (1432 bytes)` rather than the payload. Enabling DEBUG no longer starts recording sensitive content.
- **Peer log sanitizing** — control characters (notably newlines, which let a server forge additional log lines) are escaped, and messages are capped at `MAX_PEER_LOG_MESSAGE_LENGTH` (4096) with a truncation marker. Peer-chosen *severity* is intentionally still honored — that is the MCP logging utility working as designed, and the client asked for those levels via `logging/setLevel`.

### Callback error reflection (3 findings)

`csf_b2316fd0` (sampling), `csf_1aa20fb2` (SSE), `csf_2723f046` (Streamable HTTP): a host callback exception was echoed to the remote server inside the JSON-RPC error response (`"Internal error: #{e.message}"`), leaking file paths, connection strings and library internals to an untrusted peer — and giving it a probe for host internals.

Peers now receive a constant `'Internal error'` / `'Sampling error'`; the full message and backtrace stay in the local log. Error *codes* are unchanged (`-32603`, `-32601`), so spec-conformant error semantics are preserved.

## Testing

- New `spec/lib/mcp_client/sensitive_logging_spec.rb`: credentials never reach the log while non-sensitive fields do; request params and response bodies are absent while method/id/size are present.
- New logging specs: forged-newline content is escaped, oversized messages truncate.
- Three existing specs updated to assert peers get the constant message while the detail is logged locally.
- Full suite: 1613 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/client.rb
+++ b/lib/mcp_client/client.rb
@@ -29,6 +29,20 @@ module MCPClient
     # :warn logs a warning on mismatch, :strict raises a ValidationError.
     STRUCTURED_CONTENT_MODES = %i[warn strict].freeze
 
+    # Server-config keys whose values carry credentials (HTTP headers, the
+    # subprocess environment, inline tokens). Their values are replaced before
+    # a config is written to the log.
+    SENSITIVE_CONFIG_KEYS = %i[headers env token access_token api_key auth authorization
+                               password secret client_secret oauth_provider].freeze
+
+    # Placeholder written in place of a redacted value.
+    REDACTED = '[REDACTED]'
+
+    # Maximum characters of a peer-supplied log message written to the host
+    # log. The remote server controls this content, so an unbounded message
+    # would let it inflate log storage at will.
+    MAX_PEER_LOG_MESSAGE_LENGTH = 4096
+
     # Initialize a new MCPClient::Client
     # @param mcp_server_configs [Array<Hash>] configurations for MCP servers
     # @param logger [Logger, nil] optional logger, defaults to STDOUT
@@ -65,7 +79,7 @@ module MCPClient
         @logger.formatter = proc { |severity, _datetime, progname, msg| "#{severity} [#{progname}] #{msg}\n" }
       end
       @servers = mcp_server_configs.map do |config|
-        @logger.debug("Creating server with config: #{config.inspect}")
+        @logger.debug("Creating server with config: #{redact_config(config).inspect}")
         MCPClient::ServerFactory.create(config, logger: @logger)
       end
       @tool_cache = {}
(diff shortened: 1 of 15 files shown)


simonx1:

**Verdict: `needs-attention`** — the vulnerabilities confirmed real, but the fix was **incomplete in four places**. All four are now closed. This was the most useful review of the batch: I'd fixed the paths I went looking for and missed the ones I didn't.

> Config redaction and sampling-error masking work, but the security fix is incomplete.

### Finding 1 — `[high]` Elicitation handler exceptions still reflected to peers

> This rescue converts a host callback exception into an error result containing `e.message`. **Because the exception is caught here, the new constant-message transport rescues never run**; `send_elicitation_response` forwards this message over SSE, Streamable HTTP, or stdio. A branch reproduction returned a host-secret-bearing exception verbatim in the -32603 response.

Correct, and the reasoning is the part I got wrong: I hardened the *transport* rescues and assumed they covered every path, but the elicitation rescue sits upstream in the client, so it wins. Reproduced with a handler raising a connection string:

```
before: "Elicitation handler error: db://user:hunter2@10.0.0.5/prod is down"
after:  "Elicitation handler error"
```

### Finding 2 — `[high]` Both SSE transports still DEBUG-logged complete wire payloads

> `process_event_chunk` logs the raw Streamable HTTP SSE chunk before parsing, so sampling prompts, elicitation data, notifications, and replayed responses still enter lower-trust logs without a size cap. Equivalent misses remain in `server_sse.rb:845` and `server_sse/json_rpc_transport.rb:210`.

Also correct. I replaced the *JSON-RPC message* logging and missed the *raw chunk* logging that sits one layer below it — same data, earlier in the pipeline. All three now log byte counts, with a test asserting a sentinel secret in an SSE chunk never reaches the log on either transport.

### Finding 3 — `[medium]` stdio still reflected exception details

> Unlike the SSE and Streamable HTTP siblings, the stdio dispatcher still interpolates `e.message` into its peer-facing -32603 response. […] **The branch's stdio elicitation spec still explicitly expects the leaking message.**

Right on both counts — I fixed two of three transports and left a spec pinning the leak in the third. Fixed and the spec now asserts the constant message.

### Finding 4 — `[medium]` An invalid log level bypassed sanitization and the cap

> `logger_name` and `data` are sanitized, but the fallback branch interpolates the peer-controlled **level** verbatim. A `notifications/message` payload with `level="bogus\nERROR forged"` creates an injected log line, and an oversized level bypasses `MAX_PEER_LOG_MESSAGE_LENGTH`.

Correct — I sanitized two of the three peer-controlled fields in that method. The level now goes through the same sanitizer, with a newline-injection test.

### Verification

Full suite 1615 examples / 0 failures, RuboCop clean.

PR #199, opened 2026-07-30
Route task operations to the task's own server

simonx1:
## Summary

Batched security fix for three low-severity findings from the codex-security scan (`csf_879782f9` `get_task`, `csf_c09a1eab` `get_task_result`, `csf_5b68e696` `cancel_task` — `routing-error.task-server-selection`).

All three task methods resolved their target server with `select_server(server)` where a `nil` selector returns `@servers.first`. Task IDs are only unique within the server that issued them, so in a multi-server client a call like `client.get_task('task-123')` silently went to server 0 regardless of which server was actually running that task — polling an unrelated task's state, retrieving another server's result, or cancelling work that was never targeted (whichever task happened to share the ID).

## Fix

Server resolution for task operations is now explicit and fails closed:

1. an explicit `server:` argument wins;
2. a `MCPClient::Task` handle carries the server that issued it — `Task#server` was already populated by `call_tool_as_task`, it just wasn't consulted;
3. a bare ID with exactly one configured server remains unambiguous and works as before;
4. anything else raises `ArgumentError` with a message pointing at both remedies, instead of guessing.

All three methods now accept either a `Task` or a `String` ID.

**Compatibility**: single-server clients — the common case and every README/example flow — are unaffected. A multi-server client that passed a bare ID was already routing arbitrarily; it now gets a clear error instead of silent misrouting.

README and the two task examples now pass the `Task` handle (`client.get_task(task)`), which is the idiom that makes routing correct by construction.

## Testing

- New `spec/lib/mcp_client/task_server_routing_spec.rb`: handle-based routing reaches the issuing server and never the first server (for get/result/cancel); explicit `server:` overrides the handle; bare IDs raise in multi-server clients but work in single-server ones and when the server is named.
- Full suite: 1615 examples, 0 failures. RuboCop clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,15 +327,24 @@ task = client.call_tool_as_task('long_job', { input: 'data' }, ttl: 60_000)
 # honoring the server's suggested poll interval
 until task.terminal? || task.input_required?
   sleep((task.poll_interval || 1000) / 1000.0)
-  task = client.get_task(task.task_id)   # tasks/get
+  task = client.get_task(task)          # tasks/get, routed to the task's own server
 end
 
 # Retrieve the underlying result (e.g. a CallToolResult) via tasks/result
-result = client.get_task_result(task.task_id)
+result = client.get_task_result(task)
 
 # List and cancel tasks
 page = client.list_tasks               # { tasks: [...], next_cursor: ... }
-client.cancel_task(task.task_id)       # tasks/cancel
+client.cancel_task(task)               # tasks/cancel
+```
+
+Task IDs are only unique within the server that issued them, so pass the `Task`
+returned by `call_tool_as_task` — it carries its own server. A bare task ID also
+works when the client has a single server; with several servers configured it
+raises `ArgumentError` rather than guessing, so name the server explicitly:
+
+```ruby
+client.get_task('task-123', server: 'my-server')
 
 # React to server-pushed status updates
 client.on_notification do |server, method, params|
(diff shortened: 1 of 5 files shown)


simonx1:

**Verdict: `needs-attention`** — finding confirmed real, routing confirmed correct, one regression found and fixed.

### Was the finding real? Yes, and the routing works

> The vulnerability on main is real: all three task methods call `select_server(nil)`, which returns the first configured server (`lib/mcp_client/client.rb` main:571-575, 599-603, 638-643, 848-853). The diff correctly routes valid Task handles and intentionally breaks ambiguous multi-server bare-ID calls, while preserving valid single-server calls.

Useful confirmation that the deliberate breaking change lands only where intended.

### Finding — `[medium]` `server: false` dispatched instead of being rejected

> The new helper uses **truthiness** to detect an explicit selector. Consequently, `false` bypasses `select_server(false)`, which previously raised `ArgumentError`, and is treated as an omitted selector. With a Task handle it routes through `task.server`; with one configured server it selects that server. A malformed or dynamically supplied `server: false` can therefore **read or cancel a task** instead of failing locally. A runtime probe confirmed both `tasks/result` and `tasks/cancel` were issued.

**Correct, and a regression I introduced** — `if server_arg` conflates "absent" with "present but invalid". It matters more than a typical nil-vs-false nit because the affected verbs include `cancel_task`: a caller building options dynamically (`server: opts[:server] && …`) would get a silent cancel where they previously got a loud `ArgumentError`.

**Fix:** presence is now `nil`-checked, so `false` falls through to `select_server` and raises as before. Test asserts zero RPCs and an `ArgumentError` for both `cancel_task` (multi-server, with a Task handle) and `get_task` (single-server).

### Verification

Full suite 1616 examples / 0 failures, RuboCop clean.

PR #200, opened 2026-07-30
Bound SSE event ids and floor the resumption reconnect delay

simonx1:
## Summary

Batched security fix for two low-severity findings from the codex-security scan, both in the Streamable HTTP resumption path (`csf_378a2f4e` `resource-exhaustion.unbounded-event-id`, `csf_4135aed9` `resource-exhaustion.unbounded-remote-input`).

**Unbounded event id** — `parse_and_handle_event` stored every `id:` field into `@last_event_id`, which `apply_events_headers` then echoes as the `Last-Event-ID` header on every subsequent events/resumption request. A server could therefore make the client retain an arbitrarily large string indefinitely and emit oversized outbound headers — or, with control characters, a header value that is illegal in HTTP.

**Zero-delay resumption loop** — `run_resumption_loop` computed `delay = retry_ms ? retry_ms / 1000.0 : 0` and honored a peer-supplied `retry: 0` literally, so a replay server could drive back-to-back GETs with no pause for the whole read-timeout window.

## Fix

- `retainable_event_id?` gates cursor retention: non-empty, ≤ `MAX_EVENT_ID_LENGTH` (1024), and printable ASCII (`\x21-\x7E`, which notably excludes CR/LF). A rejected id is logged and **leaves the previous valid cursor intact**, so resumption keeps working from the last good position rather than losing its place.
- `resumption_delay` floors peer directives at `MIN_RESUMPTION_RECONNECT_DELAY` (10 ms). That is deliberately far smaller than the events-stream floor: SEP-1699's polling pattern wants fast reconnects, and this only prevents `retry: 0` from becoming a busy loop.

## Note for reviewers

This touches `server_streamable_http.rb` near the constants block and the resumption loop, as does #193 (events-stream retry floor). If both land, expect a small textual conflict in the constants region — the two floors are intentionally separate values with different rationales, so keep both.

## Testing

- New specs: ordinary ids still recorded; oversized and header-illegal ids ignored; previous cursor preserved across a rejected id; `retry: 0` floors while a 250 ms directive is honored exactly and a missing directive falls back to the default.
- Full suite: 1613 examples, 0 failures. RuboCop clean.

### DIFF
--- a/lib/mcp_client/server_streamable_http.rb
+++ b/lib/mcp_client/server_streamable_http.rb
@@ -33,6 +33,22 @@ module MCPClient
     SSE_MAX_RECONNECT_DELAY = 30   # Maximum reconnect delay in seconds
     THREAD_JOIN_TIMEOUT = 5 # Timeout for thread cleanup
 
+    # Floor for the delay between resumption GETs. SEP-1699's polling pattern
+    # wants fast reconnects, so this is far smaller than the events-stream
+    # floor — it only prevents a peer-supplied "retry: 0" from turning the
+    # deadline window into a back-to-back request loop.
+    MIN_RESUMPTION_RECONNECT_DELAY = 0.01
+
+    # Maximum length of a server-supplied SSE event id retained as the
+    # resumption cursor. The id is echoed in the Last-Event-ID header of
+    # subsequent requests, so an unbounded value means unbounded retained
+    # memory and oversized outbound headers.
+    MAX_EVENT_ID_LENGTH = 1024
+
+    # Characters allowed in a retained event id: printable ASCII only, since
+    # the value becomes an HTTP header. Notably excludes CR/LF.
+    EVENT_ID_PATTERN = /\A[\x21-\x7E]+\z/
+
     # @!attribute [r] base_url
     #   @return [String] The base URL of the MCP server
     # @!attribute [r] endpoint
@@ -771,7 +787,7 @@ module MCPClient
       state = { cursor: cursor, retry_ms: retry_ms }
       # SEP-1699: the client MUST respect the server's retry directive before
       # attempting to reconnect.
-      delay = retry_ms ? retry_ms / 1000.0 : 0
+      delay = resumption_delay(retry_ms)
 
       loop do
         sleep(delay) if delay.positive?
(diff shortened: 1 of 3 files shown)


simonx1:

**Verdict: `needs-attention`** — findings confirmed real, three problems raised. All three fixed.

> Main's vulnerabilities were real: IDs were stored unconditionally and emitted as headers […] and `retry:0` skipped every sleep in the resumption loop.

### Finding 1 — `[high]` Three of four id paths bypassed the validator

> The POST SSE path still assigns every non-empty event ID directly to `@last_event_id`. `resume_or_fail` also passes that raw per-stream ID to the resumption GET […] Separately, `handle_resumption_event` assigns raw IDs to `state[:cursor]` […] Neither path calls `retainable_event_id?`. A runtime probe on HEAD **retained and reflected a 5,000-character POST ID** and advanced a resumed-GET cursor to the same value.

**Correct** — I validated the GET events parser and missed that ids reach `Last-Event-ID` through three other routes. The cap was real but trivially sidestepped by using a different stream.

All four sites now share `retainable_event_id?`. Fixing this also surfaced that the validator lived on the class while most callers are in the `JsonRpcTransport` module — it's now in the module, so every includer (and the module's own spec double) has it.

### Finding 2 — `[medium]` The first resumption GET was newly delayed by 1 s

> Main used a zero delay for the **first** resumption GET when `retry_ms` was nil, applying `SSE_RECONNECT_DELAY` only between later attempts. The new initial call to `resumption_delay(nil)` returns one second. This adds latency to every no-directive resumption and can consume the entire read timeout; with `read_timeout: 1`, the queue timeout can kill the worker before useful I/O and turn a previously resumable response into the non-retryable `ServerError` path.

Correct, and an unintended behaviour change from routing the initial delay through the same helper: the *floor* was the point, not a *new* default wait. The first attempt is immediate again when no directive was given; the 10 ms floor still applies to an explicit `retry: 0`, and the 1 s fallback still applies between later attempts.

### Finding 3 — `[medium]` The charset rejected valid ids containing spaces

> The allowed range starts at 0x21, excluding ASCII space (0x20), even though space is printable ASCII and legal inside an HTTP field value. […] an SSE ID such as `"cursor 42"` is rejected and the old cursor remains. Reconnection can therefore use a **stale cursor**, causing duplicate replay or failed/missed resumption with otherwise valid servers.

Fair — I reused a session-id-style range without rechecking it against HTTP field-value rules, and the failure mode is worse than rejecting the id: resumption silently continues from an older position. Range is now `0x20-0x7E`; CR/LF remain excluded, which is the part that matters.

### Verification

New tests cover an oversized id on the POST path and the resumed-GET path, an id with an internal space being accepted, and the first no-directive resumption GET going out without sleeping. Full suite 1617 examples / 0 failures, RuboCop clean.

PR #201, opened 2026-07-30
Harden examples: sandbox roots, pinned npm servers, no auto tool calls

simonx1:
## Summary

Batched security fix for **twelve** low-severity findings from the codex-security scan, all in `examples/`. They share one root cause: the examples hand a model a set of MCP tools and execute whatever it selects, unattended, in a process holding live API credentials.

### Confused deputy — blast radius (6 findings)

`csf_a53a8d0f` (Anthropic), `csf_6e7bee6c` (OpenAI), `csf_cc319de1` (Gemini), `csf_6f0ec6a0` (Streamable HTTP), `csf_d4db8113` (RubyLLM), `csf_a2004a0e` (ruby-openai).

- The three filesystem examples rooted the server at `Dir.pwd` — the whole checkout, including `.git` and `examples/secrets.env` (live tokens, when the runner is configured). They now use a **disposable sandbox** (`examples/support/example_sandbox.rb`): a tmp directory seeded with sample files, removed at exit, and the run prints what it can reach. The Gemini case was the sharpest — that process can also read the Vertex service-account file.
- `streamable_http_example.rb` searched the server's advertised tools for any zero-argument one and **called it**. Against Zapier that is an arbitrary connected action (send mail, move money, change records) chosen by the server, not the user. It now *names* a candidate and calls nothing unless you set `MCP_EXAMPLE_TOOL`.
- The two browser examples document that they execute every model-selected Playwright action without confirmation, and that page content read by one tool can influence the next call — so they need a disposable browser profile, not an authenticated one.

### Supply chain (5 findings)

`csf_36298d0e`, `csf_834092af`, `csf_bd31f1ff` (unversioned `server-filesystem`), `csf_3f21ac26`, `csf_750fc629` (`@playwright/mcp@latest`).

`npx -y` resolves and executes a mutable release with no integrity check, in processes that hold `ANTHROPIC_API_KEY` / `OPENAI_API_KEY` / Vertex credentials — and `run_all_examples.sh` sources `secrets.env` into the environment of everything it launches. Both packages are now pinned (`@modelcontextprotocol/server-filesystem@2026.7.10`, `@playwright/mcp@0.0.78` — both verified to resolve on the registry), overridable via `MCP_FS_SERVER_VERSION` / `PLAYWRIGHT_MCP_VERSION`, across the examples, the runner script, and `sample_server_definition.json`.

### Plaintext credential storage (1 finding)

`csf_1ec706e6`: the `FileTokenStorage` demo wrote access tokens, refresh tokens and client secrets via `File.write`, leaving the mode to the process umask (commonly `0644`, world-readable). It now creates the file `0600` and says plainly that production storage belongs in an OS keychain or secrets manager.

## Testing

- Every modified file syntax-checked (`ruby -c`, `bash -n`, JSON parse); RuboCop clean across `examples/`.
- Sandbox helper smoke-tested: creates the tmp root with sample files and emits the pinned, sandbox-scoped command.
- Pinned versions confirmed present on the npm registry (HTTP 200).
- Library suite unaffected: 1613 examples, 0 failures.

Note: the LLM examples themselves make paid API calls and were not executed here.

### DIFF
--- a/examples/gemini_ai_mcp.rb
+++ b/examples/gemini_ai_mcp.rb
@@ -4,7 +4,7 @@
 # MCPClient integration example using the `gemini-ai` Ruby gem
 #
 # Prerequisites
-#  - A running MCP server (for instance: `npx @modelcontextprotocol/server-filesystem .`)
+#  - A running MCP server (for instance: `npx @modelcontextprotocol/server-filesystem@2026.7.10 .`)
 #  - A Vertex AI service-account JSON file (downloaded from Google Cloud) named
 #    `google-credentials.json` in the project root, or set its location in
 #      export VERTEX_CREDENTIALS_FILE=/path/to/file.json
@@ -20,6 +20,7 @@
 
 require 'bundler/setup'
 require_relative '../lib/mcp_client'
+require_relative 'support/example_sandbox'
 
 # The gem itself
 require 'gemini-ai'
@@ -52,9 +53,13 @@ end
 logger = Logger.new($stdout)
 logger.level = Logger::WARN
 
-# Connect using stdio - passing an array of command arguments auto-detects stdio transport
+# Connect using stdio - passing an array of command arguments auto-detects stdio transport.
+# The server is pinned to a known version and rooted at a disposable sandbox, not
+# the checkout: this process can read the Vertex credentials file, and the model
+# below picks the tool calls that run unattended.
+ExampleSandbox.announce_auto_execution("filesystem tools under #{ExampleSandbox.filesystem_root}")
 mcp_client = MCPClient.connect(
-  %W[npx -y @modelcontextprotocol/server-filesystem #{Dir.pwd}],
+  ExampleSandbox.filesystem_server_command,
   logger: logger
 )
 
@@ -195,7 +200,7 @@ second_input = {
     function_declarations: google_tools
   },
   contents: [
-    { role: 'user', parts: { text: 'List all files in current directory' } },
+    { role: 'user', parts: { text: 'List all files in the sandbox directory' } },
     candidate['content'], # the assistant message containing the functionCall
     { role: 'function', parts: [function_response_part] }
   ],
--- a/examples/gemini_ai_mcp.rb
+++ b/examples/gemini_ai_mcp.rb
@@ -60,6 +60,7 @@ logger.level = Logger::WARN
 ExampleSandbox.announce_auto_execution("filesystem tools under #{ExampleSandbox.filesystem_root}")
 mcp_client = MCPClient.connect(
   ExampleSandbox.filesystem_server_command,
+  env: ExampleSandbox.secret_free_env,
   logger: logger
 )
 
--- a/examples/json_input_mcp_servers_example.rb
+++ b/examples/json_input_mcp_servers_example.rb
@@ -6,7 +6,7 @@
 # - Filesystem MCP server via STDIO
 #
 # Usage:
-#  1. Start Playwright MCP server: npx @playwright/mcp@latest --port 8931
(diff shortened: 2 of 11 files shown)


simonx1:

I asked Codex (`codex-cli 0.146.0`, adversarial review, branch diff vs `main`) to independently confirm the findings and check the fixes.

**Verdict: `needs-attention`** — the original defects confirmed real, three gaps found in the fix. All three are now closed.

> The defects were real on main: model-driven servers exposed `Dir.pwd` (`examples/openai_ruby_mcp.rb:25`, `examples/ruby_anthropic_mcp.rb:22`, `examples/gemini_ai_mcp.rb:57`), streamable HTTP invoked an arbitrary zero-argument tool (`examples/streamable_http_example.rb:58-65`), and OAuth storage used umask-dependent `File.write` (`examples/oauth_example.rb:156`).

### Finding 1 — `[high]` Pinning the top-level package doesn't stop credential exfiltration

> This pins only the top-level filesystem package before executing it through npx. The published 2026.7.10 package has no shrinkwrap and declares **ranged runtime dependencies** including `@modelcontextprotocol/sdk ^1.29.0`, `diff ^8.0.3`, `glob ^10.5.0`, and `minimatch ^10.0.1`, so a later compromised matching dependency can still be installed and executed silently. MCPClient's Open3 subprocess **inherits the parent environment**, including `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` and the Vertex credential path.

**Correct — I'd pinned the front door and left the environment wide open.** The pin narrows *which* code runs; it does nothing about *what that code can steal*.

**Fix:** `ExampleSandbox.secret_free_env` passes every credential-looking variable to `Open3` as `nil`, which unsets it in the child, and `run_all_examples.sh` strips the same set before launching Playwright (it sources `secrets.env` into its own environment, so it was the worst offender). Verified against my own shell — which turned out to be carrying `BRAVE_API_KEY`, `CLAUDE_API_KEY` and others:

```
child sees: openai=[] zapier=[] vertex=[] example=[ok]
```

The transitive-pin half of the recommendation (a committed lockfile with integrity hashes) I have **not** done — that means vendoring an npm lockfile into `examples/`, which is a larger call for a demo directory. Stripping the environment removes the payoff, which is the part that matters here; happy to add the lockfile if you'd like it.

### Finding 2 — `[medium]` Explicit tool requests could "succeed" without running

> `MCP_EXAMPLE_TOOL` accepts any advertised name but always invokes it with `{}`, so tools with required arguments fail. A typo or stale tool name merely prints a warning, continues to ping, and emits "Example completed successfully." Exceptions are also rescued without a nonzero exit.

Right, and that regression was mine — the opt-in contract I introduced was silently unreliable. A missing or argument-requiring tool now raises, and the example exits non-zero whenever anything failed, so CI and scripts can tell.

### Finding 3 — `[medium]` The 0600 mode only applied to *new* files

> The 0600 creation mode applies only when the file is new. For an `oauth_tokens.json` created by the old code as 0644, `File.open` preserves that mode, truncates and writes refreshed tokens, closes the file, and only then chmods it. Another local user can read during that interval, and a crash or chmod failure leaves the rewritten credentials world-readable indefinitely — the exact upgrade path this hardening needs to secure.

A good catch: the upgrade path is precisely the case the change existed for, and it was the one case it didn't cover. Writes now go to a fresh `0600` temp file in the same directory, fsync, then `rename(2)` over the destination. Verified starting from a 0644 file:

```
before: 644
after:  600
content preserved: ["https://api.example.com"]
temp cleaned up: true
```

### Verification

RuboCop clean across `examples/`, all example files parse, library suite 1606 examples / 0 failures.

PR #202, opened 2026-07-30
Scope example servers to their session and fix fixture path containment

simonx1:
## Summary

Batched security fix for **seven** low-severity findings from the codex-security scan, all in the demo/fixture servers. Each one is a missing authorization or containment check that a real server must not copy from an example.

### Task ownership — `examples/echo_server_streamable.py` (4 findings)

`csf_1051a2ae` (cancel), `csf_f1baa2cb` (get), `csf_d2eb058f` (list), `csf_cefeb09c` (result).

`tasks_store` was process-global and every handler looked a task up by ID alone. Any connected session could therefore poll another session's task state, retrieve its completed result, cancel its in-progress work, or enumerate the entire process-wide inventory via `tasks/list`.

Tasks now record `_owner_session_id` at creation. `tasks/get`, `tasks/result` and `tasks/cancel` treat another session's task exactly like a missing one — so IDs can't be probed for existence — and `tasks/list` returns only the caller's own tasks.

### Session binding — `examples/elicitation/elicitation_streamable_server.py` (1 finding)

`csf_bd31ed78`. A POST whose `Mcp-Session-Id` header was missing or unknown fell through to "find the first active session". With more than one client connected, that silently bound one client's requests to another client's session — including elicitation *responses* carrying whatever the user typed — and queued the replies onto that other client's SSE stream.

The `endpoint` event now advertises `/sse?sessionId=<id>` (a same-origin relative URI, per MCP 2024-11-05), so every POST identifies its own session. A POST with no session id gets `400`; an unknown/inactive one gets `404`. No guessing.

### Path containment — `spec/support/fake_filesystem_mcp_server.rb` (2 findings)

`csf_6f8d9c0c` (lexical prefix), `csf_62a6556f` (symlink). Containment was `target.to_s.start_with?(ROOT_DIR.to_s)` on an unresolved path, which accepts a sibling directory whose name merely begins with the root (`/srv/root-backup` vs `/srv/root`) and follows a symlink straight out of the root. Now both sides are `realpath`-resolved (falling back to the deepest existing ancestor for paths that don't exist yet) and compared per path component.

## Testing

Verified by running the servers, not just by reading them:

- **New spec** `spec/support/fake_filesystem_mcp_server_spec.rb` drives the fixture over stdio: both escapes reproduced as failures before the fix and pass after; in-root listing and `..` traversal still behave.
- **Cross-session probe** against the running echo server: session A creates a task, session B gets `TaskNotFound` for get and cancel and an empty `tasks/list`, while A still sees `working` and lists its own. `examples/tasks_example.rb` completes its full lifecycle unchanged.
- **Elicitation SSE round trip** (`test_elicitation_sse_simple.rb`) completes against the modified server; a POST with no session id returns 400 and one with an unknown id returns 404.

### DIFF
--- a/examples/echo_server_streamable.py
+++ b/examples/echo_server_streamable.py
@@ -74,7 +74,7 @@ def _now_iso():
     return datetime.now(timezone.utc).isoformat().replace("+00:00", "Z")
 
 
-def create_task_record(task_id, tool_name, arguments, ttl=None):
+def create_task_record(task_id, tool_name, arguments, ttl=None, owner_session_id=None):
     """Create a task record in the MCP 2025-11-25 shape."""
     now = _now_iso()
     task = {
@@ -86,6 +86,7 @@ def create_task_record(task_id, tool_name, arguments, ttl=None):
         "ttl": ttl,
         "pollInterval": 300,
         # Internal bookkeeping (not sent to the client)
+        "_owner_session_id": owner_session_id,
         "_tool_name": tool_name,
         "_arguments": arguments,
         "_result": None,
@@ -93,6 +94,18 @@ def create_task_record(task_id, tool_name, arguments, ttl=None):
     with tasks_lock:
         tasks_store[task_id] = task
     return task
+
+
+def task_owned_by(task, session_id):
+    """Whether a task belongs to the requesting MCP session.
+
+    Task IDs are handed to one session, so every task operation must be scoped
+    to its owner: otherwise any connected client could poll, read the result
+    of, or cancel another session's task by guessing or observing its ID.
+    Tasks created before ownership was recorded have no owner and stay
+    reachable only by a session-less caller.
+    """
+    return task.get("_owner_session_id") == session_id
 
 
 def task_to_response(task):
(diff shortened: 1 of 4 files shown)


simonx1:

I asked Codex (`codex-cli 0.146.0`, adversarial review, branch diff vs `main`) to independently confirm the findings and check the fixes.

**Verdict: `needs-attention`** — all findings confirmed real, the SSE and filesystem fixes confirmed complete, one bypass found in the task isolation. Fixed.

> No-ship: all reported vulnerabilities are real on main. The SSE and filesystem fixes close their stated escapes, but task isolation remains bypassable for session-less callers.

### Finding — `[medium]` Session-less callers still shared task ownership

> The route reads an **optional** session header without requiring a valid active session (lines 256-257), then records that value as the owner (lines 899-900). `task_owned_by` treats equality as authorization **even when both values are `None`**. Consequently, two independent no-header callers share one anonymous owner: I reproduced one caller creating a task and a second caller successfully listing, getting, and cancelling it.

**Correct, and it's the classic shape of this bug** — I added an owner field and an equality check without asking what happens when the owner is absent. `None == None` is true, so every unidentified caller became the owner of every unidentified task, and my two-session probe missed it because both of my clients *did* have sessions.

**Fix:** `tasks/*` and task-augmented `tools/call` now require a live session (`-32600` otherwise), and `task_owned_by` treats an absent owner *or* an absent requester as unauthorized rather than as a match.

Verified against the running server:

```
anonymous tasks/list   -> {"code": -32600, "message": "Missing or unknown Mcp-Session-Id for task operation"}
anonymous tasks/cancel -> {"code": -32600, ...}
```

and the legitimate paths still work — `examples/tasks_example.rb` completes its full lifecycle, and the two-session probe still shows B getting `TaskNotFound` for A's task and an empty `tasks/list` while A sees its own.

### Verification

Library suite 1610 examples / 0 failures, RuboCop clean, Python servers compile.

PR #203, opened 2026-07-30
Pin GitHub Actions to commit SHAs

simonx1:
## Summary

Security fix for a low-severity finding from the codex-security scan (`csf_07420c4b`, `supply-chain.mutable-github-action`).

`.github/workflows/ci.yml` referenced third-party actions by mutable tag (`@v1`, `@v4`). A tag is a movable pointer: the action's owner — or anyone who compromises that account — can repoint it at new code, and every subsequent workflow run picks it up silently. The `code-review` job is the sharp end, since it grants the referenced action `secrets.ANTHROPIC_API_KEY`, `pull-requests: write`, and `id-token: write` (OIDC minting).

## Fix

All three actions are pinned to the full commit SHA their tag currently resolves to, with the version in a trailing comment:

| Action | Pinned SHA | Was |
|---|---|---|
| `actions/checkout` | `11d5960a326750d5838078e36cf38b85af677262` | `@v4` |
| `ruby/setup-ruby` | `95ef2b042f9d7a56d8268cba8559e2842e2ad01b` | `@v1` |
| `anthropics/claude-code-action` | `be7b93b1907a4abad570368f3c74b6fe3807510b` | `@v1` |

Each SHA was resolved from the tag through the GitHub API and verified to exist in the upstream repository. This is GitHub's own [recommended hardening](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions), and Dependabot understands the `SHA # version` form, so update PRs still arrive — they just become reviewable diffs instead of invisible tag moves.

## Note on the code-review check

The `code-review` job fails on most PRs in this repo (~370 ms, `is_error:true`, `$0` cost), including before this change and on re-runs. It happened to succeed on this PR's first run, but a same-day re-run of an unpinned branch failed again with the identical signature — so this is flaky/expiring upstream configuration (likely the `ANTHROPIC_API_KEY` secret or the pinned `claude-opus-4-6` model), **not** something pinning fixed or broke. The `test` jobs are the reliable signal. Worth a separate look.

Note also that `v1` here is an *annotated* tag: its ref object dereferences to commit `be7b93b…`, which is what this PR pins. Pinning the commit (not the tag object) is the correct form.

## Testing

Workflow YAML parses and every `uses:` resolves to the intended pinned reference. No library code touched.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,10 @@
 name: CI
 
+# Third-party actions are pinned to a full commit SHA rather than a mutable
+# tag. A tag can be repointed at new code by the action's owner (or anyone who
+# compromises that account), and here it would inherit ANTHROPIC_API_KEY,
+# pull-requests: write, and OIDC-minting rights. The trailing comment records
+# the human-readable version the SHA corresponds to; bump both together.
 on:
   push:
     branches: [ main ]
@@ -14,10 +19,10 @@ jobs:
         ruby-version: ['3.2', '3.3']
 
     steps:
-    - uses: actions/checkout@v4
-    
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+
     - name: Set up Ruby
-      uses: ruby/setup-ruby@v1
+      uses: ruby/setup-ruby@95ef2b042f9d7a56d8268cba8559e2842e2ad01b # v1
       with:
         ruby-version: ${{ matrix.ruby-version }}
         bundler-cache: true
@@ -36,10 +41,10 @@ jobs:
       pull-requests: write
       id-token: write
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
       with:
         fetch-depth: 0
-    - uses: anthropics/claude-code-action@v1
+    - uses: anthropics/claude-code-action@be7b93b1907a4abad570368f3c74b6fe3807510b # v1
       with:
         prompt: |
           REPO: ${{ github.repository }}


simonx1:

**Verdict: `needs-attention`** — finding confirmed real, the pins independently verified correct, and one residual supply-chain path raised that this PR does not close.

### The pins check out

> The pre-change finding was real: main used mutable aliases at `.github/workflows/ci.yml:17-20` and `:39-42` while the review job held write/OIDC permissions and the API key at `:34-49`. The new SHAs correctly map to [checkout v4.4.0](https://github.com/actions/checkout/releases/tag/v4.4.0), [setup-ruby v1.321.0](https://github.com/ruby/setup-ruby/releases/tag/v1.321.0), and [claude-code-action v1.0.183](https://github.com/anthropics/claude-code-action/releases/tag/v1.0.183); **actionlint passes and no expression-based command injection was found**.

That's the check I most wanted from an independent pass — a pinned SHA is worthless if it points somewhere unintended, and I'd resolved them through the API rather than eyeballing releases. It also ran actionlint and looked for template-injection sinks, which this PR doesn't touch but is the other classic hazard in a workflow holding `pull-requests: write`.

### Finding — `[high]` The pinned action still runs an unpinned `curl | bash` installer

> Line 47 freezes the GitHub repository revision, but that exact release still executes `curl -fsSL https://claude.ai/install.sh | bash -s -- 2.1.220`. The installer runs **after** the action places `GITHUB_TOKEN`/`GH_TOKEN` in its environment and inherits the `ANTHROPIC_API_KEY` […] Thus compromise or mutation of the installer endpoint remains a direct arbitrary-code path to the API key, pull-request write token, and OIDC capability. The reported mutable-GitHub-tag vector is fixed, but the same high-impact supply-chain outcome remains possible through this selected pinned action.

**I think this is accurate and worth knowing, and I'm deliberately not fixing it here.** Reasoning:

- It is a property of the third-party action's *internals*, not of this diff. Pinning the action was the reported finding and is fully addressed; this is a second, independent supply-chain surface that pinning was never going to close.
- The suggested remedy (install the Claude CLI from a digest-verified artifact and pass `path_to_claude_code_executable` so the action skips `install.sh`) means owning and maintaining a vendored binary or a verification step in this repo. That's a real decision about how much of a vendor's toolchain you want to take responsibility for — your call, not something to slip into a one-line pinning PR.
- Worth weighing against context: the `code-review` job only runs on same-repo PRs from non-Dependabot actors (line 33), which limits who can trigger it.

If you do want it closed, the shape is: fetch a pinned Claude CLI release, verify its digest, then point the action at it — happy to do that as a follow-up.

### Verification

No code change in response to the review. Workflow YAML parses, all `uses:` resolve to the intended pinned refs, and CI is green on this branch.

<sub>Codex review requested and summarized by Claude Code; the quoted text is Codex's, the assessment is mine.</sub>

PR #208, opened 2026-08-04
Release 2.1.0 — Hostile-Server Hardening

simonx1:
## Summary

Releases **2.1.0**, covering the PRs merged since 2.0.0 (#188–#210): a security pass over every transport, plus Ruby 4.0.6 support.

> **Rebased.** Codex's review of this PR found two real code bugs. Rather than smuggle them into a release commit, they shipped separately as **#209** (`tools/call` replayed by session recovery) and **#210** (peer payloads reaching the SSE error logs), both merged. This branch was rebuilt on top of them, so its diff is now version + docs + the two doc-driven fixes below — no security code changes. The review comment below has the reproductions.

The theme is one sentence: **a remote MCP server is untrusted input.** 2.0.0 was correct against a cooperative server but assumed good faith in places where a hostile — or merely compromised — peer controls the data. The wire protocol is unchanged and the ordinary client API is unchanged; what changed is what the client accepts, retries, logs and reflects back.

## What's in the release

| | |
|---|---|
| Version | `lib/mcp_client/version.rb`, `Gemfile.lock` self-reference → **2.1.0** (gemspec reads the constant) |
| CHANGELOG | New 2.1.0 entry in the established format, with **Breaking Changes** and **Migration notes** |
| README | New "Treating the Server as Untrusted" section; `tools/call` retry semantics; `max_decompressed_body_bytes` |

### Behaviour changes hosts must act on

These are spelled out in the CHANGELOG's Migration notes:

- **`tools/call` is never retried automatically** — including through session-expiry recovery (#209). A "transient" failure can arrive after the server executed the request, and JSON-RPC has no idempotency key to make a replay safe. Retry explicitly if your application knows it is safe, and treat the error as *outcome unknown*.
- **Task operations refuse to guess a server.** Pass the `Task` handle from `call_tool_as_task`, or name the server. A bare ID still works with one configured server; with several it raises `ArgumentError` rather than acting on the wrong one.
- **Peer-facing error messages are constant** and **DEBUG logs no longer contain payloads** — if you parsed either for detail, the detail now lives only in local logs. Error codes are unchanged.
- **Cross-origin SSE endpoints/redirects and non-HTTPS or loopback OAuth discovery URLs from a peer are refused.** Local development still works when the *configured* server is itself local.

## Two defects found while writing the docs

Writing the documentation turned out to be a review in its own right:

1. **`max_decompressed_body_bytes` could not be set through the documented path.** `streamable_http_config` did not accept it and `ServerFactory` did not forward it, so the option introduced in #188 only worked when constructing `ServerStreamableHTTP` directly — i.e. the knob was effectively unreachable for anyone using the normal config API. Wired through both, with specs asserting the config path, the default, and the positive-integer validation.
2. **The README claimed "No runtime dependencies"** while the gemspec declares `faraday`, `faraday-follow_redirects`, `faraday-retry` and `base64`. Pre-existing, corrected here.

## RuboCop config

`Metrics/ParameterLists` now sets `CountKeywordArgs: false`. The `*_config` builders are keyword-only option factories where each keyword is a documented, defaulted setting; counting them against a positional-argument limit would push new options into an opaque hash. The codebase already carried **six hand-written `rubocop:disable Metrics/ParameterLists`** comments for exactly this reason — those are now removed, which is what the 58 autocorrections in the diff are.

## Verification

| Check | Result |
|---|---|
| RSpec, Ruby 4.0.6 (default) | **1698 examples, 0 failures** |
| RSpec, Ruby 3.3.5 | **1698 examples, 0 failures** |
| RuboCop | 136 files, no offenses |
| `gem build` | builds `ruby-mcp-client-2.1.0.gem`; ships lib + LICENSE + README only (no specs/examples/secrets) |
| `Gem::Specification#validate` | passes |
| Version consistency | `version.rb`, `Gemfile.lock`, gemspec all report 2.1.0 |
| Examples suite (pre-release, on merged main) | 18 PASS · 0 FAIL · 2 SKIP |

The two example skips are environmental: no `ANTHROPIC_API_KEY` in this environment, and the interactive browser-OAuth example.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -35,6 +35,11 @@ Metrics/PerceivedComplexity:
 
 Metrics/ParameterLists:
   Max: 8
+  # The *_config builders are keyword-only option factories: every keyword is a
+  # documented, defaulted transport setting. Counting those against a limit
+  # meant for positional arguments would push new options into an opaque
+  # options hash, which is worse for callers.
+  CountKeywordArgs: false
 
 Metrics/ModuleLength:
   Max: 500
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 A security pass over every transport, driven by an external scan of the 2.0.0
 codebase (50 findings: 12 medium, 38 low) and a second, adversarial review of
-each fix (PRs #188–#210). Every finding was reproduced before it was fixed and
-re-verified afterwards — including two bugs found by reviewing the release
-notes themselves against the code (#209, #210).
+each fix (PRs #188–#211). Every finding was reproduced before it was fixed and
+re-verified afterwards — including three bugs found by reviewing the release
+notes themselves against the code (#209, #210, #211).
 
 The theme: **a remote MCP server is untrusted input.** 2.0.0 was correct against
 a cooperative server but assumed good faith in places where a hostile — or merely
@@ -45,15 +45,18 @@ retries, logs and reflects back.
   server. Peers receive `'Internal error'` / `'Sampling error'` /
   `'Elicitation handler error'`; the detail stays in the local log. Error
   **codes** are unchanged.
-- **Logs no longer contain payloads** (#198, #210). Request params, response
-  bodies and raw SSE chunks are replaced by a method/id summary and a byte count,
-  so enabling DEBUG no longer records `tools/call` arguments, tool results or
-  elicitation content. The SSE *error* paths are covered too: a non-object JSON
-  payload used to be logged with `#inspect` at **WARN**, which the default
-  logger emits, so that one leaked without DEBUG being enabled at all. Server
-  configs are logged with credential-bearing keys redacted, and peer-supplied
-  log messages are control-character escaped and capped (a server could
-  otherwise forge log lines).
+- **Logs no longer contain payloads** (#198, #210, #211). Request params,
+  response bodies and raw SSE chunks are replaced by a method/id summary and a
+  byte count, so enabling DEBUG no longer records `tools/call` arguments, tool
+  results or elicitation content. The *error* paths are covered too, and they
+  were the leakier ones: a non-object JSON payload used to be logged with
+  `#inspect` at **WARN** (which the default logger emits, so it leaked with
+  DEBUG off), and `JSON::ParserError#message` quotes the offending token, so
+  malformed peer JSON echoed into logs and into the `Invalid JSON response from
+  server` exception on every transport. Parse failures now report position and
+  size only. Server configs are logged with credential-bearing keys redacted,
+  and peer-supplied log messages are control-character escaped and capped (a
+  server could otherwise forge log lines).
 - **Cross-origin traffic is refused on the legacy SSE transport** (#189). An
   `endpoint` control event that changes scheme/host/port fails the handshake, and
   redirects that leave the connection origin are refused — `faraday-follow_redirects`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,150 @@
 # Changelog
 
+## 2.1.0 — Hostile-Server Hardening (2026-08-04)
+
+A security pass over every transport, driven by an external scan of the 2.0.0
+codebase (50 findings: 12 medium, 38 low) and a second, adversarial review of
+each fix (PRs #188–#210). Every finding was reproduced before it was fixed and
+re-verified afterwards — including two bugs found by reviewing the release
+notes themselves against the code (#209, #210).
+
+The theme: **a remote MCP server is untrusted input.** 2.0.0 was correct against
+a cooperative server but assumed good faith in places where a hostile — or merely
+compromised — peer controls the data. Nothing here changes the wire protocol, and
+the ordinary client API is unchanged; what changes is what the client accepts,
+retries, logs and reflects back.
+
+### Breaking Changes
+
+- **`tools/call` is never retried automatically** (#196). A "transient" failure
+  (HTTP 5xx, dropped connection) can arrive *after* the server executed the
+  request, so replaying it risks a duplicate side effect, and JSON-RPC has no
+  idempotency key to make that safe. `NON_IDEMPOTENT_METHODS` is excluded from
+  `with_retry` on all four transports. Idempotent methods (`tools/list`,
+  `resources/read`, `ping`, …) retry exactly as before. Hosts that relied on
+  tool calls being retried must now retry explicitly and decide for themselves
+  whether re-execution is acceptable.
+
+  This also covers **session-expiry recovery** on the HTTP transports (#209).
+  A 404 carrying an expired `Mcp-Session-Id` still starts a fresh session, but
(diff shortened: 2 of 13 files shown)


simonx1:

I asked Codex (`codex-cli 0.146.0`, adversarial review, branch diff vs `main`) to review this release with a release reviewer's priorities: version completeness, **documentation accuracy**, the new option's plumbing, packaging, and the RuboCop change.

**Verdict: `needs-attention`.** The mechanics were clean — but it found **four documented security guarantees the code did not actually honor**. That is the most valuable thing it could have caught here: a release note is a promise, and three of these were promises I wrote confidently.

> No ship. Version metadata, gemspec validation, packaged files, RuboCop, direct builder/factory plumbing, and 1,695 tests pass, but the release documentation makes four materially false security guarantees.

All four are now fixed — two by changing code, two by narrowing claims. Each was reproduced before I touched anything.

### 1 — `[high]` `tools/call` *was* still replayed, via session recovery

> The claimed invariant is bypassed by `HttpTransportBase`: any session-bearing HTTP 404 invokes `restart_session_and_resend`, which resends the exact request without consulting `NON_IDEMPOTENT_METHODS` (`lib/mcp_client/http_transport_base.rb:237-284`). A runtime probe with a `tools/call` request produced two identical POSTs.

**Correct, and the most important find in this whole security effort.** #196 closed the retry path and I documented an absolute — while a second, independent replay path sat one layer away in session recovery. My reproduction:

```
before: tools/call POSTs: 2  ["initialize","notifications/initialized","tools/call",
                              "initialize","notifications/initialized","tools/call"]
after:  tools/call POSTs: 1  → MCPClient::Errors::ConnectionError
```

A 404 usually means the request was rejected — but it doesn't *prove* that, since a session can expire after the tool ran, which is exactly the ambiguity #196 exists for. The session is still restarted (recovery is worth keeping); the request is only re-sent when idempotent, otherwise `ConnectionError` — chosen because `with_retry` never rescues it, so no other path can turn it into a second attempt. Two regression tests: `tools/call` attempted once with the session still restarted, and `tools/list` still recovering normally.

### 2 — `[high]` Payloads *were* still written to logs

> Streamable HTTP logs valid non-object SSE payloads using `message.inspect` at WARN (`.../json_rpc_transport.rb:255-264`) and malformed event data verbatim at DEBUG (`server_streamable_http.rb:1088-1095`). A runtime probe confirmed a scalar containing `patient-record-456` appeared in the WARN log.

Also correct, and worse than a DEBUG leak because WARN is on by default. I audited the *summary* paths in #198 and missed the *error* paths — the places that log precisely because something was unexpected. Both now log type/size only, with a test asserting sentinel values never appear.

That reproduction also surfaced a small robustness bug: a JSON-parseable scalar on the events stream used to raise inside `dispatch_server_message` (caught upstream, but sloppy). It's now skipped with a typed warning.

### 3 — `[medium]` The OAuth SSRF claim was broader than the check

> The release says a remote server can no longer aim discovery locally, but validation checks only the hostname text. The implementation explicitly acknowledges that DNS names resolving to private addresses are not caught (`oauth_provider.rb:652-666`).

Fair — the *code* comment was honest about this and my *release note* was not. I chose not to implement resolution-time filtering: doing it properly means re-validating every connection and redirect and pinning the resolved address, which is a feature with real TOCTOU subtleties, not a release-week edit. Instead the claim now says "literal" loopback/private address and names the gap explicitly, with a pointer to network-layer egress control.

### 4 — `[medium]` The untrusted-server table overstated coverage

> This table is introduced as behavior of "the transports" and presents compressed responses as universally bounded. In fact, `max_decompressed_body_bytes` exists only in Streamable HTTP; plain HTTP has no size limit.

Correct. I confirmed only Streamable HTTP sends `Accept-Encoding: gzip` (`server_streamable_http.rb:140`) and plain HTTP's `parse_response` has no limit. The row is now labelled Streamable HTTP-only rather than implying blanket coverage.

### What it confirmed as sound

Version metadata consistent, gemspec valid, packaged files correct, RuboCop clean, the `max_decompressed_body_bytes` plumbing working through the documented path, and 1,695 tests passing.

### Verification after the fixes

Full suite **1698 examples, 0 failures** on Ruby 4.0.6 and 3.3.5; RuboCop clean; both original reproductions now report the documented behaviour.

simonx1:
## Codex follow-up verification (round 2)

I asked Codex to re-review after the restructure, checking three things: whether its four earlier findings are actually resolved **in the working tree** (two fixes now live in `main`, not in this diff), whether the split is clean, and whether the rewrite introduced any new inaccuracy.

**Verdict: `needs-attention`** — three of four resolved, split confirmed clean, and **one finding was not actually fixed**.

> Findings #1, #3, and #4 are resolved, the release split is clean, and 1,698 specs plus RuboCop pass. Finding #2 remains exploitable because malformed peer JSON still reaches default-visible logs.

### Confirmed resolved

| Prior finding | Status |
|---|---|
| #1 `tools/call` replayed via session recovery | resolved (#209) |
| #3 OAuth SSRF claim overstated | resolved — docs narrowed |
| #4 response-size table overstated | resolved — row labelled Streamable HTTP only |
| Split clean / release content only | confirmed |

### Finding #2 was **not** fixed — and my test said it was

> The fix assumes `JSON::ParserError#message` contains only a position, but Ruby includes the offending token. I reproduced a POST SSE event containing `LEAKME` producing a WARN log containing `LEAKME`; the GET events path […] similarly logged `SECONDLEAK` at ERROR, and legacy SSE has the same pattern at `server_sse/sse_parser.rb:45`. **The regression test passes accidentally because its secret occurs after the first invalid token.**

Both halves of that are correct, and the second half is the more useful one.

I had replaced the log lines that printed payloads directly but kept `e.message`, believing it carried only a position. It doesn't:

```ruby
JSON.parse('{LEAKME-123-45-6789 bad}')
# => expected object key, got 'LEAKME-123-45-6789' at line 1 column 2
```

So peer bytes still reached logs — on **five** paths, not the two I'd looked at: the GET events stream (ERROR), the POST SSE and legacy SSE parsers (**WARN, default-visible**), the OAuth token-refresh warning (where the payload is a *token response*), and the `Invalid JSON response from server` exception raised by all four transports.

And the test I wrote to prove the fix **passed by accident** — its sentinel sat after the first invalid token, so the parser never quoted it. A green test that cannot fail is worse than no test, because it stops anyone looking again.

**Fixed in #211** (merged): `describe_parse_error` keeps the `line N column M` position and the payload size — the parts that make a broken server diagnosable — and drops the quoted content. New tests place the sentinel as the **first** invalid token on each path, which is the case that actually exercises the quoting.

### This PR now

Rebased on `main` (which carries #209, #210, #211). Its diff is version + docs + the two doc-driven fixes; `git diff main` shows zero changes in the security-critical files. The CHANGELOG credits #211 and the "logs no longer contain payloads" entry now describes the parser-message leak explicitly rather than implying blanket coverage.

Full suite **1701 examples, 0 failures**; RuboCop clean; CI green on 4.0.6 / 3.3 / 3.2.

